PR #8, opened 2015-08-10
Fix unit tests for hadoop-plugin module and fix how cross-plugin task…

convexquad:
… dependencies are added


convexquad:
+1 Self-approving so we can pass PCL and publish a new version of the li-hadoop-plugin multiproduct

PR #9, opened 2015-08-11
LIHADOOP-12945 Hadoop Plugin azkabanUpload enhancement to print proje…

convexquad:
…ct URL and reset version to 0.5.x series


convexquad:
+1 Self-approving commit that was committed to the internal LinkedIn gitli repo, but not to the Github repo. I think this was the only one that got missed.

Resetting the version to 0.5.17 so we can publish a final build in the 0.5.x series before we go on to 0.6.x.

PR #10, opened 2015-08-11
Enable Hadoop Plugin tests to work wth LinkedIn mint snapshot builds

convexquad:
Previously when you ran a LinkedIn "mint snapshot" build the hadoop-plugin-test module unit tests failed, because putting -SNAPSHOT in the version number is a LinkedIn-specific construct. This will enable mint snapshot builds to succeed, which helps in local testing at LinkedIn.


convexquad:
+1 Self-approving as I am about to follow up with commits to the LinkedIn multiproduct wrapper so we can publish a good 0.6.x version.

PR #11, opened 2015-08-12
Oozie upload task

nntnag17:
JIRAs fixed: 
LIHADOOP-13353 Hadoop Plugin oozieUpload task should upload to versioned directory.
LIHADOOP-13457 Fix Stackoverflow exception thrown by OozieUploadTask.

### DIFF
--- a/hadoop-plugin/src/main/groovy/com/linkedin/gradle/hdfs/HdfsFileSystem.groovy
+++ b/hadoop-plugin/src/main/groovy/com/linkedin/gradle/hdfs/HdfsFileSystem.groovy
@@ -197,7 +197,25 @@ class HdfsFileSystem {
    * @param p path to check
    * @return true if p exists else returns false
    */
-  public Boolean exists(Path p) {
-    fs.exists(p);
+  public boolean exists(Path p) {
+    return fs.exists(p);
+  }
+
+  /**
+   * @param p the path to delete
+   * @param recursive true if delete recursively
+   * @return true if successful else false
+   */
+  public boolean delete(Path p, Boolean recursive) {
+    logger.info("deleting ${p.toString()}")
+    return fs.delete(p,recursive);
+  }
+
+  /**
+   * @param p the path to delete
+   * @return true if successful else false
+   */
+  public boolean delete(Path p) {
+    return delete(p,true);
   }
 }
--- a/hadoop-plugin/src/main/groovy/com/linkedin/gradle/hdfs/HdfsFileSystem.groovy
+++ b/hadoop-plugin/src/main/groovy/com/linkedin/gradle/hdfs/HdfsFileSystem.groovy
@@ -192,4 +192,12 @@ class HdfsFileSystem {
     throws IOException {
     fs.copyFromLocalFile(delSrc, overwrite, src, dst);
   }
+
+  /**
+   * @param p path to check
+   * @return true if p exists else returns false
+   */
+  public Boolean exists(Path p) {
+    fs.exists(p);
+  }
 }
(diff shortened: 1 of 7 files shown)


hadoop-plugin/src/main/groovy/com/linkedin/gradle/oozie/OozieUploadTask.groovy
```diff
@@ -21,11 +21,14 @@ class OozieUploadTask extends DefaultTask {
     fs.initialize(getURI());
 
     Path directoryPath = new Path(oozieProject.dirToUpload);
-    Path projectPath = new Path(oozieProject.uploadPath + oozieProject.projectName);
+    Path projectPath = new Path(oozieProject.uploadPath + oozieProject.projectName + "/v${getProject().version}");
 
     try {
       logger.info("Project path: ${projectPath.toString()}");
       logger.info("Directory path: ${directoryPath.toString()}");
+      if(fs.exists(projectPath)) {
```
convexquad: Ah, we need to be able to oozieUpload multiple times and have it overwrite the previous version. At least that's what I think right now - I'm not completely sure yet.  Can you cause it to overwrite? I think we want to keep the exists() method you added, even if we end up not using it.

li-hadoop-plugin/src/main/groovy/com/linkedin/gradle/liutil/LiKerberosUtil.groovy
```diff
@@ -8,12 +8,13 @@ class LiKerberosUtil {
 
   static Logger logger = LoggerFactory.getLogger(LiKerberosUtil.class);
   private static final String OOZIE_KRB5_CONF = "oozie-krb5.conf";
+  private static final String JAVA_TMP_DIR = System.getProperty("java.io.tmpdir");
 
   /**
    * We need to copy the oozie-krb5.conf file to a temporary location since we cannot point property java.security.krb5.conf to the file inside jar.
    */
   static void writeConfToTemp() {
-    logger.debug("writing oozie-krb5.conf file to ${System.getProperty("java.io.tmpdir")}");
```
convexquad: Is there something in particular that made you have to change this? I think it's good to get the temp dir at the time the method is used, rather than as a static class variable, even if this means you might call System.getProperty("java.io.tmpdir") more than once. Generally, getting things like temp directory locations at static classloading times is kind of a bad thing.

convexquad:
I commented on a couple of issues - let's resolve them, and then I'll +1 this.

nntnag17:
I've updated the pull request addressing your comments. The copyFromLocalFile method takes a boolean 'overwrite' to overwrite files but it doesn't work while copying directories. Therefore as a work around to this, I'm first deleting the hdfs directory if it exists and then calling `mkdir` and 'copyFromLocalfile`

hadoop-plugin/src/main/groovy/com/linkedin/gradle/hdfs/HdfsFileSystem.groovy
```diff
@@ -192,4 +192,30 @@ class HdfsFileSystem {
     throws IOException {
     fs.copyFromLocalFile(delSrc, overwrite, src, dst);
   }
+
+  /**
+   * @param p path to check
+   * @return true if p exists else returns false
+   */
+  public Boolean exists(Path p) {
+    fs.exists(p);
```
convexquad: Can you add the "return" keyword to these methods, even though it's not required in Groovy?
Please make these return little boolean instead of Boolean

convexquad:
I added one more comment. +1 for you to merge this PR after fixing that comment.

nntnag17:
I've fixed it. Merging the PR.

PR #12, opened 2015-08-13
Update the li-hadoop-plugin override of the writeAzkabanPluginJson me…

convexquad:
…thod to fill in some common property values


convexquad:
This PR should make it easier for people to use the azkabanUpload task, since some of the properties will already start with ok values in the .azkabanPlugin.json value. @akshayrai, can you review this Pull Request?

PR #15, opened 2015-08-19
LIHADOOP-13660 Hadoop Plugin oozieUpload task should upload the conte…

nntnag17:
…nts of the hadoopZip task.

### DIFF
--- a/hadoop-plugin/src/main/groovy/com/linkedin/gradle/hdfs/HdfsFileSystem.groovy
+++ b/hadoop-plugin/src/main/groovy/com/linkedin/gradle/hdfs/HdfsFileSystem.groovy
@@ -21,10 +21,14 @@ import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.hdfs.web.WebHdfsFileSystem;
+import org.apache.hadoop.io.IOUtils;
 import org.apache.hadoop.security.UserGroupInformation;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipInputStream;
 
 /**
  * HDFS file system abstraction for the Hadoop Plugin that uses WebHDFS.
@@ -100,7 +104,7 @@ class HdfsFileSystem {
 
     def process = processBuilder.start();
     process.waitFor();
-    return (process.exitValue == 0);
+    return (process.exitValue() == 0);
   }
 
   /**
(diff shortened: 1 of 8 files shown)


hadoop-plugin/src/main/groovy/com/linkedin/gradle/hdfs/HdfsFileSystem.groovy
```diff
@@ -250,4 +295,37 @@ class HdfsFileSystem {
...
+   * @param zipIn Inputstream from a zip file
+   * @param filePath File path where zipIn should be written
+   * @throws IOException If zipIn cannot be written to filePath
+   */
+  private extractFile(ZipInputStream zipIn, String filePath) throws IOException {
+
+    // get output stream from the filesystem
+    OutputStream out = fs.create(new Path(filePath));
+
+    InputStream is = zipIn;
```
convexquad: Is this upcast actually necessary?

hadoop-plugin/src/main/groovy/com/linkedin/gradle/hdfs/HdfsFileSystem.groovy
```diff
@@ -250,4 +295,37 @@ class HdfsFileSystem {
...
+    // get the name of the file
+    String fileName = filePath.substring(filePath.lastIndexOf("/")+1);
+
+    logger.info("copying ${fileName} to ${filePath}");
+
+    try {
+      // copy the file to the hdfs
+      IOUtils.copyBytes(is, out, conf, false);
+    }
+    catch(Exception e) {
```
convexquad: Remove the catch block - just let the exception propagate up the stack! It should just have try-finally.

hadoop-plugin/src/main/groovy/com/linkedin/gradle/hdfs/HdfsFileSystem.groovy
```diff
@@ -192,6 +196,47 @@ class HdfsFileSystem {
...
+        // if the entry is a directory, create it
+        mkdir(new Path(filePath));
+      }
+      // close this zip entry
+      zipIn.closeEntry();
+      // get the next entry
+      entry = zipIn.getNextEntry();
+    }
+    // close input stream once the zip is extracted
+    zipIn.close();
```
convexquad: I think we need a try-finally block so that we are sure to close zipIn

hadoop-plugin/src/main/groovy/com/linkedin/gradle/hdfs/HdfsFileSystem.groovy
```diff
@@ -250,4 +295,37 @@ class HdfsFileSystem {
...
+    InputStream is = zipIn;
+
+    // get the name of the file
+    String fileName = filePath.substring(filePath.lastIndexOf("/")+1);
+
+    logger.info("copying ${fileName} to ${filePath}");
+
+    try {
+      // copy the file to the hdfs
+      IOUtils.copyBytes(is, out, conf, false);
```
convexquad: Question 1: will this use the Kerberos information? Can you double check that it fails when you have not kinited, and then succeeds once you do?
Question 2: does it use webhdfs (since that is what we have set in the conf)?

hadoop-plugin/src/main/groovy/com/linkedin/gradle/hdfs/HdfsFileSystem.groovy
```diff
@@ -192,6 +196,47 @@ class HdfsFileSystem {
...
+
+    logger.info("zip file path: ${zipFile.toString()}");
+    logger.info("destination directory: ${destDirectory.toString()}");
+
+    ZipInputStream zipIn = new ZipInputStream(new FileInputStream(zipFile));
+    ZipEntry entry = zipIn.getNextEntry();
+
+    //iterates over entries in the zip file
+    while (entry != null) {
+      String filePath = destDirectory.toString() + File.separator + entry.getName();
```
convexquad: Does entry.getName() include the path information in the zip? Let's say that in the zip the entry has the path /src/main/pig/myScript.pig. On HDFS, does it get copied to ${destDirectory}/src/main/pig/myScript.pig or just to ${destDirectory}/myScript.pig? I hope this preserves the path information (we must find a way to preserve it).

Super glad you have found a way to do this without extracting all the files from the zip into a temp directory - I really hope this approach allows us to preserve the path information in the zip as we copy it to the dest directory on HDFS!

hadoop-plugin/src/main/groovy/com/linkedin/gradle/hdfs/HdfsFileSystem.groovy
nntnag17: I've verified kerberos authentication. It works.

It uses webhdfs since we are using the webhdfs object(fs) that we have created while initializing this HdfsFileSystem. I think this is also the reason why Kerberos authentication works.

hadoop-plugin/src/main/groovy/com/linkedin/gradle/hdfs/HdfsFileSystem.groovy
nntnag17: entry.getName gives us the name of the entry, if it is a file it will give the filename and similar for directory. We differentiate between directory and files in the zip by calling entry.isDirectory() method. When the entry is a directory, we create the directory in hdfs and when the entry is a file, we extract the file using extractFile method. This way the path information is preserved. I've tested this by uploading a zip containing various nested directories and the whole directory structure was extracted to the hdfs.

nntnag17:
@convexquad I've updated the pull request with your comments

convexquad:
+1

PR #18, opened 2015-08-25
The oozieUpload task should depend on the buildHadoopZips task but do…

convexquad:
…es not seem to do so. Also update the NOTICE file.


convexquad:
+1 Self-approving

PR #19, opened 2015-09-04
LIHADOOP-14099 Create spark job type in hadoop plugin dsl.

nntnag17:
Spark job type in hadoop plugin dsl.


hadoop-plugin/src/main/groovy/com/linkedin/gradle/hadoopdsl/job/SparkJob.groovy
```diff
@@ -0,0 +1,182 @@
...
+
+  /**
+   * DSL enableFlags method sets the flags for the job. When the job file is build, job properties
+   * of the form flag.flagName=true is added in the job file.
+   * @param flags
+   */
+  void enableFlags(List<String> flags) {
+    this.flags = flags;
+    flags.each { flag ->
+      setJobProperty("flag.$flag",'true');
```
johnyu0520: maybe we can discuss real quick.
the current implementation is I'll look for the flag in the props, if it's there, it is going to be true, otherwise it is going to be false.

Given that, setJobProperty(flag.$flag, "false") is still going to be true.  Wondering what's a good way to not mislead the users?

nntnag17: Hey @johnyu0520 ,

The method setJobProperty is never exposed to the user. Users can specify the flags which should be true by enableFlags method. 
    `def flags = ['version','verbose','help']`
    `enableFlags flags`

When a **flagname** is in the list of specified flags. flag.**flagname**=true is written to the job file. There is no way users can write flag.**flagname**=false in the job file unless they edit it manually, which is not recommended.

convexquad:
Just FYI I went out of town on vacation and will be in transit on Monday - I'll definitely review this PR on Tuesday!!!

hadoop-plugin/src/main/groovy/com/linkedin/gradle/hadoopdsl/job/SparkJob.groovy
akshayrai: Hi @johnyu0520,
I think having ```flag.<flag>=enable/disable``` would be easier to understand than having ```flag.<flag>="doesn't matter"``` in the Azkaban Spark jobtype.

If flag is enabled then we set it else ignore it(default behaviour).

nntnag17:
Updated the PR

*Addressed Akshay's comments.
*Updated VERSIONS.md, CONTRIBUTORS.md
*Updated few comments.

hadoop-plugin/src/main/groovy/com/linkedin/gradle/hadoopdsl/job/SparkJob.groovy
johnyu0520: Hey @akshayrai, @nntnag17 
Thanks for the comments.  What you guys say make sense.  So just to confirm again, I'll change the behavior to the following:

if ( flag.<flag> == "true" ) 
  //flag enabled
else
  //flag disabled

Note I'll specifically be looking for the word "true" and nothing else

hadoop-plugin/src/main/groovy/com/linkedin/gradle/hadoopdsl/HadoopDslFactory.groovy
```diff
@@ -160,6 +161,15 @@ class HadoopDslFactory {
...
 
   /**
+   * Factory method to build a SparkJob
+   *
+   * @param name
+   * @return The job
+   */
+  SparkJob makeSparkJob(String name) {
+    return new SparkJob(name);
+  }
```
convexquad: Can you add a space and add the JavaDoc comment for the name parameter

hadoop-plugin/src/main/groovy/com/linkedin/gradle/hadoopdsl/job/SparkJob.groovy
```diff
@@ -0,0 +1,182 @@
...
+ *   def flags = ['version','verbose','help']
+ *   def params = ['param1','param2']
+ *
+ *   sparkJob('jobName') {
+ *     uses 'com.linkedin.azkaban.spark.HelloSpark'  // Required
+ *     executes './lib/azkaban-spark-job.jar' // Required
+ *     appParams params // application parameters
+ *     enableFlags flags
+ *
+ *     set Properties: [
```
convexquad: "set properties:" here with a lowercase p

convexquad: I'm confused about "set Properties:" - is it supposed to be the same as the regular "set properties:" method? If so, can you fix the documentation here to use the lowercase "properties:" since it's case-sensitive. If it's something different, can we make it more descriptive, like "set sparkProperties:" or something like that?

hadoop-plugin/src/main/groovy/com/linkedin/gradle/hadoopdsl/job/SparkJob.groovy
```diff
@@ -0,0 +1,182 @@
...
+ *       'key1' : 'val1',
+ *       'key2' : 'val2'
+ *     ]
+ *
+ *   }
+ * </pre>
+ */
+
+class SparkJob extends HadoopJavaProcessJob {
+  Map<String, Object> properties;
```
convexquad: Nothing seems to add to the member variable "properties" - see my comment above about "set Properties:", I'm guessing these issues are related.

convexquad:
I left a couple of minor comments, but I have one more request - update the test files:

hadoop-plugin-test/src/main/gradle/positive/jobs1.gradle
hadoop-plugin-test/src/main/gradle/negative/missingFields.gradle

and the resulting expectedJobs / expectedOutput files under hadoop-plugin-test/expectedJobs and hadoop-plugin-test/expectedOutput (you should have to add a file to the directory hadoop-plugin-test/expectedJobs/jobs1 and update hadoop-plugin-test/expectedOutput/negative/missingFields.out).

Other that that, this PR looks pretty good from the Hadoop DSL perspective.

hadoop-plugin/src/main/groovy/com/linkedin/gradle/hadoopdsl/job/SparkJob.groovy
```diff
@@ -0,0 +1,182 @@
...
+   *
+   * @param args Args whose key 'properties' has a map value specifying the job properties to set;
+   *   or a key 'jvmProperties' with a map value that specifies the JVM properties to set;
+   *   or a key 'confProperties' with a map value that specifies the Hadoop job configuration properties to set;
+   *   or a key 'confs' with a map value that specifies the Spark configuration to set
+   */
+  @Override
+  void set(Map args) {
+    super.set(args);
+    if (args.containsKey("confs")) {
```
convexquad: At the Hadoop DSL level, I would be in favor of naming this something just a little more descriptive - maybe "set sparkConfs:"? Is that better than "set confs:"???

hadoop-plugin/src/main/groovy/com/linkedin/gradle/hadoopdsl/job/SparkJob.groovy
nntnag17: This is same as "set properties". I'll make the correction.

hadoop-plugin/src/main/groovy/com/linkedin/gradle/hadoopdsl/job/SparkJob.groovy
nntnag17: Renamed to sparkConfs.

hadoop-plugin/src/main/groovy/com/linkedin/gradle/hadoopdsl/job/SparkJob.groovy
nntnag17: I'll remove it. Initially I was thinking of adding properties using "set properties" but then I realized that this functionality is already inherited from the "job" class.

nntnag17:
Addressed Alex and Ratandeep's comments.

convexquad:
+1 Looks great

johnyu0520:
Looks good from the communication with the Spark job perspective

PR #21, opened 2015-09-22
Dsl jobs

nntnag17:


nntnag17:
Updated the PR with support for property files.

nntnag17:
Updated PR with LIHADOOP-14506: Hadoop plugin should directly submit jobs to oozie.

convexquad:
@nntnag17 thank you for submitting all the ELR tickets! I reopened one as one jar still isn't in Artifactory. Let's hold off merging this PR request until Darius finishes with the last one.

convexquad:
Tentative +1 once the last jar gets submitted to Artifactory. Note for the public: these are LinkedIn internal systems issues.

nntnag17:
@convexquad The last jar is not actually required. I'd removed it in the last commit and that's why didn't create ELR ticket for it.  I'll address the remaining comments and update the PR today.

convexquad:
@nntnag17 After you update the PR, can you run "mint wc-test" to launch a PCX run? It shouldn't complain about jar issues. If it fails for some other reason, it's ok, you are still +1 to merge. If it fails due to missing dependencies, let's hold off on merging until we have verified we have all the necessary jars in Artifactory.

PR #24, opened 2015-10-14
LIHADOOP-15077 Hadoop plugin should have a feature to control and mon…

nntnag17:
…itor dependencies specified by the users

### DIFF
--- a/hadoop-plugin/src/main/groovy/com/linkedin/gradle/dependency/CheckDependencyTask.groovy
+++ b/hadoop-plugin/src/main/groovy/com/linkedin/gradle/dependency/CheckDependencyTask.groovy
@@ -1,0 +1,212 @@
+/*
+ * Copyright 2015 LinkedIn Corp.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.linkedin.gradle.dependency;
+
+import groovy.json.JsonSlurper;
+import org.gradle.api.DefaultTask;
+import org.gradle.api.Project;
+import org.gradle.api.artifacts.Dependency;
+import org.gradle.api.tasks.TaskAction;
+
+class CheckDependencyTask extends DefaultTask {
+
+  String DEPENDENCY_PATTERN_FILE = "sample-dependency-pattern.json";
+  boolean errorFound = false;
+  boolean warningFound = false;
+
+
+  @TaskAction
+  void checkDependencies() {
+
+    Map<Dependency, List<DependencyPattern>> matchedDependencyPatterns;
+
+    // get all the dependencies used in the project
+    Set<Dependency> dependencies = getDependencies(project);
+    project.logger.debug("Dependencies found : ${dependencies.toString()}")
+
+    // get the List of DependencyPattern against which the dependencies should be checked
+    List<DependencyPattern> dependencyPatterns = getDependencyPatterns(project);
+
+    project.logger.debug("DependencyPatterns found : ${dependencyPatterns.toString()}")
+
+    matchedDependencyPatterns = fillMatchedDependencyPatterns(dependencies, dependencyPatterns);
+
+    matchedDependencyPatterns.each {
+      dependency, matchedPatterns ->
+        matchedPatterns.each {  matchedPattern -> handleDependencyCheck(project, dependency, matchedPattern) }
+    }
+  }
+
+  /**
+   * Fills matchedDependencyPatterns
+   * @param dependencies The set of dependencies
+   * @param dependencyPatterns The list of dependency patterns.
+   */
+  Map<Dependency, List<DependencyPattern>> fillMatchedDependencyPatterns(Set<Dependency> dependencies, List<DependencyPattern> dependencyPatterns) {
+
+    Map<Dependency, List<DependencyPattern>> matchedDependencyPatterns = new HashMap<Dependency, List<DependencyPattern>>();
+    dependencies.each {
+      dependency ->
+        dependencyPatterns.each {
+          dependencyPattern ->
+            // if any dependency matches dependencyPattern then put it in matchedDependencyPatterns
+            if (dependencyMatchesPattern(project, dependency, dependencyPattern)) {
+              if (!matchedDependencyPatterns.containsKey(dependency)) {
+                matchedDependencyPatterns.put(dependency, new LinkedList<DependencyPattern>());
+              }
+              matchedDependencyPatterns.get(dependency).add(dependencyPattern);
+            }
+        }
+    }
+    return matchedDependencyPatterns;
+  }
+
+  /**
+   * This method handles the check of dependencies and calls other handlers based on the severity
+   * @param project The Gradle project
+   * @param dependency The dependency to check
+   * @param matchedPattern The matchedPattern against which dependency should be checked
+   */
+  void handleDependencyCheck(Project project, Dependency dependency, DependencyPattern matchedPattern) {
+    switch (matchedPattern.severity) {
+      case SEVERITY.ERROR:
+        errorFound = true;
+        handleError(project, dependency, matchedPattern);
+        break;
+      case [SEVERITY.INFO, null]:
+        handleInfo(project, dependency, matchedPattern);
+        break;
+      case SEVERITY.WARN:
+        warningFound = true;
+        handleWarn(project, dependency, matchedPattern);
+        break;
+    }
+  }
+
+  /**
+   * Method to handle error. Subclasses can override this method to handle error in their own way.
+   * @param project The gradle project
+   * @param dependency The dependency to check
+   * @param matchedPattern The matchedPattern against which dependency should be checked
+   */
+  void handleError(Project project, Dependency dependency, DependencyPattern matchedPattern) {
+    project.logger.error("ERROR: ($dependency.group:$dependency.name:$dependency.version), ${matchedPattern.getMessage()}");
+  }
+
+  /**
+   * Method to handle info. Subclasses can override this method to handle info in their own way.
+   * @param project The gradle project
+   * @param dependency The dependency to check
+   * @param matchedPattern The matchedPattern against which dependency should be checked
+   */
+  void handleInfo(Project project, Dependency dependency, DependencyPattern matchedPattern) {
+    project.logger.lifecycle("INFO: ($dependency.group:$dependency.name:$dependency.version), ${matchedPattern.getMessage()}");
+  }
+
+  /**
+   * Method to handle warning. Subclasses can override this method to handle warning in their own way.
+   * @param project The Gradle project
+   * @param dependency The dependency to check
+   * @param matchedPattern The matchedPattern against which dependency should be checked
+   */
+  void handleWarn(Project project, Dependency dependency, DependencyPattern matchedPattern) {
+    project.logger.warn("WARN: ($dependency.group:$dependency.name:$dependency.version), ${matchedPattern.getMessage()}");
+  }
+
+/**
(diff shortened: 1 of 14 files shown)


hadoop-plugin/src/main/resources/sample-dependency-pattern.xml
```diff
@@ -0,0 +1,32 @@
...
+  License for the specific language governing permissions and limitations under
+  the License.
+-->
+
+
+<!-- This is a sample dependencyPattern file. All the dependencies of the project will be checked against the regular
+expression defined in <group>, <name> and <version>. Note that the dependencyPattern will be checked if all the <group>,
+<name> and <version> matches the patterns. The severity can take values ERROR, WARN and INFO and the message
+to go with the severity can be defined in the <message> tag. -->
+<dependencies>
```
rdsr: Why have it in XML . It is extremely verbose. Why not a form of gradle notation.

hadoop-plugin/src/main/groovy/com/linkedin/gradle/dependency/CheckDependencyTask.groovy
```diff
@@ -0,0 +1,202 @@
...
+import org.gradle.api.artifacts.Dependency;
+import org.gradle.api.tasks.TaskAction;
+
+import javax.xml.bind.JAXBContext;
+import javax.xml.bind.Unmarshaller;
+
+class CheckDependencyTask extends DefaultTask {
+
+  Map<Dependency, List<DependencyPattern>> matchedDependencyPatterns;
+  String DEPENDENCY_PATTERN_FILE = "sample-dependency-pattern.xml";
```
rdsr: I feel a file with gradle notations or something similar is much easier to specify

hadoop-plugin/src/main/groovy/com/linkedin/gradle/dependency/CheckDependencyTask.groovy
```diff
@@ -0,0 +1,202 @@
...
+import org.gradle.api.Project;
+import org.gradle.api.artifacts.Dependency;
+import org.gradle.api.tasks.TaskAction;
+
+import javax.xml.bind.JAXBContext;
+import javax.xml.bind.Unmarshaller;
+
+class CheckDependencyTask extends DefaultTask {
+
+  Map<Dependency, List<DependencyPattern>> matchedDependencyPatterns;
```
rdsr: do u need this as a field?

hadoop-plugin/src/main/groovy/com/linkedin/gradle/dependency/CheckDependencyTask.groovy
```diff
@@ -0,0 +1,202 @@
...
+  void checkDependencies() {
+
+    matchedDependencyPatterns = new HashMap<Dependency, List<DependencyPattern>>();
+
+    // get all the dependencies used in the project
+    Set<Dependency> dependencies = getDependencies(project);
+    project.logger.debug("Dependencies found : ${dependencies.toString()}")
+
+    // get the List of DependencyPattern against which the dependencies should be checked
+    List<DependencyPattern> dependencyPatternList = getDependencyPatterns(project).getDependencyPattern();
```
rdsr: rename to dependencyPatterns
Also the API is not clear.
The method  "getDependencyPatterns(project).getDependencyPattern()" suggests that  it return a single dependency pattern, but it is actually returning a list. Can you rename that the reflect the correct intention

hadoop-plugin/src/main/groovy/com/linkedin/gradle/dependency/CheckDependencyTask.groovy
```diff
@@ -0,0 +1,202 @@
...
+
+    project.logger.debug("DependencyPatterns found : ${dependencyPatternList.toString()}")
+
+    fillMatchedDependencyPatterns(dependencies,dependencyPatternList);
```
rdsr: This method can return the Map and we can have it as a local variable instead of having the Map as a field

hadoop-plugin/src/main/groovy/com/linkedin/gradle/dependency/CheckDependencyTask.groovy
```diff
@@ -0,0 +1,202 @@
...
+/**
+ * Returns true if the dependency matches with dependencyPattern
+ * @param dependency The dependency to match
+ * @param dependencyPattern The dependencyPattern against which dependency should be matched
+ * @return true or false based on whether match is found or not
+ */
+  Boolean dependencyMatchesPattern(Project project, Dependency dependency, DependencyPattern dependencyPattern) {
+
+    // If all of the group, name and version are not specified then return false
+    if (dependency.group == null && dependency.name == null && dependency.version == null) {
```
rdsr: Will if ever be the case where all these 3 are null?

hadoop-plugin/src/main/groovy/com/linkedin/gradle/dependency/CheckDependencyTask.groovy
```diff
@@ -0,0 +1,202 @@
...
+  /**
+   * Returns the set of all the dependencies which should be checked.
+   * Subclasses can override this method to return their own set of dependencies. By default we
+   * return all the dependencies.
+   * @param project The gradle project
+   * @return all dependencies of the project
+   */
+  Set<Dependency> getDependencies(Project project) {
+    Set<Dependency> dependencies = new HashSet<Dependency>();
+    project.getConfigurations().each {
```
rdsr: You should provide a mechanism to configure the configurations. In gradle projects people create many private configurations for testing etc. These should not be pulled in.

hadoop-plugin/src/main/groovy/com/linkedin/gradle/dependency/CheckDependencyTask.groovy
```diff
@@ -0,0 +1,202 @@
...
+    return dependencies;
+  }
+
+  /**
+   * Creates and returns instance of DependencyPatterns class from dependencyXmlFile
+   * @param project The gradle project
+   * @return an instance of DependencyPatterns
+   */
+  DependencyPatterns getDependencyPatterns(Project project) {
+    JAXBContext jaxbContext = JAXBContext.newInstance(DependencyPatterns.class);
```
rdsr: With gradle XML seems to verbose.
Some ideas. 
Why not specify a map literal?  Something like [version: *, group: com.linkedin.*]  . You won't need to parse XML, it will read by Groovy as a Map. You'll not need to define marshaller unmarshaller classes. etc.

hadoop-plugin/src/main/groovy/com/linkedin/gradle/dependency/DependencyPattern.groovy
```diff
@@ -0,0 +1,165 @@
...
+ *                 &lt;/sequence>
+ *             &lt;/content>
+ *           &lt;/type>
+ *   &lt;/content>
+ * &lt;/type>
+ * </pre>
+ *
+ *
+ */
+@XmlAccessorType(XmlAccessType.FIELD)
```
rdsr: You won't need this verbose class if you'd allow specifying the dependency pattern as a groovy map literal, 

http://stackoverflow.com/questions/2212606/groovy-isnt-there-a-stringtomap-out-of-the-box

hadoop-plugin/src/main/groovy/com/linkedin/gradle/dependency/DependencyPatterns.groovy
```diff
@@ -0,0 +1,67 @@
...
+ *       &lt;sequence maxOccurs="unbounded" minOccurs="0">
+ *         &lt;element name="DependencyPattern"/>
+ *       &lt;/sequence>
+ *   &lt;/content>
+ * &lt;/type>
+ * </pre>
+ *
+ *
+ */
+@XmlRootElement( name= "dependencies")
```
rdsr: See above comment

hadoop-plugin/src/main/resources/sample-dependency-pattern.xml
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
```
rdsr: See above comment about not using XML

nntnag17:
Updated the PR after addressing review comments

convexquad:
This is generally a pretty cool feature and I am in favor of it. First, I would like to make a check and see if there are any other existing Gradle Plugins that the Hadoop Plugin can apply itself to solve this problem (possible), or if any standard Gradle mechanisms exist that the Hadoop Plugin can use to declare a list of blacklisted dependencies (doubtful). If if some plugin exists that tries to solve this problem, we might want to have our own version, but let's document the alternatives here on the PR and give a short reason for our decision.

@nntnag17 can you take a look at https://plugins.gradle.org, the documentation for the latest version of Gradle, and especially the Netflix Nebula Blacklist plugin (https://github.com/nebula-plugins/gradle-blacklist-plugin) and give your thoughts about them?

convexquad:
It looks good. However, I wonder if we should add something equivalent to the translate functionality in the blacklist, where we try to replace dependencies if possible?

Part of me thinks that could be a bad idea, as the end user won't have a good understanding of they end up with different dependencies in their Hadoop zips than what they declared in their product-spec / build.gradle.

convexquad:
Ok, providing functionality equivalent to the translate function in the Nebula blacklist plugin is indeed a bad idea. Users won't be able to understand why they end up with different dependencies from what they declared. Even worse, "gradle dependencies" doesn't work correctly with dependencies that get added programmatically. Instead of trying to do magic, we should just report an error when users use things listed in the dependency file as errors.

convexquad:
The last thing I'm thinking about is that if we should expose things through extension methods similar to the blacklist plugin, or if we keep the dependency file that's in the PR.

nntnag17:
I think this functionality would be used on an organizational level. I'm not sure how exposing an extension would be a better alternative if this functionality is not usable to individual users. Also, if we have  a dependency file, then adding more dependencies to block would be as easy as adding another entry in the file without changing the code. I feel having a dependency file is a better alternative. What are your thoughts?

hadoop-plugin/src/main/groovy/com/linkedin/gradle/dependency/DependencyPlugin.groovy
```diff
@@ -0,0 +1,56 @@
...
+import org.gradle.api.Task;
+
+class DependencyPlugin implements Plugin<Project> {
+
+  // by default disable dependency check. Subclasses can set this variable to true to enable dependencyCheck
+  boolean enableDependencyCheck = false;
+
+  @Override
+  void apply(Project project) {
+    if (enableDependencyCheck) {
```
convexquad: Can you add a standard "if (project.hasProperty("disableDependencyPlugin")) { return; }" to the start of the apply method? Also, I think we always want to create the checkDependencies task - although only LiHadoopPlugin should cause things to actually depend on this task automatically (I think you have it setup that way).

hadoop-plugin/src/main/groovy/com/linkedin/gradle/dependency/CheckDependencyTask.groovy
```diff
@@ -0,0 +1,202 @@
...
+  }
+
+  /**
+   * Creates and returns list of DependencyPatterns
+   * @param project The gradle project
+   * @return List of dependency patterns
+   */
+  List<DependencyPattern> getDependencyPatterns(Project project) {
+    List<DependencyPattern> dependencyPatterns = new ArrayList<DependencyPattern>();
+    String dependencyPatternFile = getDependencyPatternFile(project);
```
convexquad: Can you break out a separate helper method that gets the jsonTextToParse, maybe "getDependencyPatternsJson" or something like that? That way, subclasses could override it if they wanted to customize how the patterns file is loaded, perhaps looking for a local file first instead of looking for an embedded resource file first.

convexquad:
I opened a couple of new comments - you are +1 to merge after fixing them!

hadoop-plugin/src/main/groovy/com/linkedin/gradle/dependency/DependencyPlugin.groovy
nntnag17: I think we should not allow users to disable the dependencyPlugin by using -PdisableDependencyPlugin.  If we allow this, then users can always bypass the dependency check by disabling this plugin.

convexquad:
+1 to merge your updated PR

nntnag17:
Merged the PR.

PR #26, opened 2015-10-30
LIHADOOP-15501 checkDependencies task fails if any of groupID, name o…

nntnag17:
…r version is null


convexquad:
+1

convexquad:
@nntnag17 I merged my last PR. Be sure to rebase on that. We can leave the version number as 0.6.10 and then trigger a LinkedIn PCX run after your merge.

PR #27, opened 2015-10-30
Cleanup the implementation of scope lookup

convexquad:
Cleanup the implementation of scope lookup, including the following:
* Hadoop DSL lookup method throws an exception if the name is not found.
* Good unit tests for NamedScope class.
* NamedScope class properly distinguishes between unqualified, qualified and fully-qualified name lookups and has cleaned up algorithms for each.
* NamedScope lookup method returns an instance that contains references to the declaring scope, the bound value, and the fully-qualified bound name. This allows clients of the NamedScope class to disambiguate names bound to the null value and names that are unbound.


convexquad:
Reviewers: I'll review / merge my own PR here, since this is a (small) update to the core Hadoop DSL language functionality.

convexquad:
I made some cleanups to the PR, and now I'm happy with it: +1

PR #28, opened 2015-11-03
Expose lookupRef as a DSL method

convexquad:


convexquad:
+1 Merging - safe and has good unit test coverage

PR #29, opened 2015-11-05
Make LinkedIn subclass for the Spark Plugin use eat1-nertzgw03 instea…

convexquad:
…d of eat1-nertzgw01


convexquad:
+1 trivial fix

PR #30, opened 2015-11-09
LIHADOOP-15728 Update dependency pattern blacklist rules

nntnag17:


li-hadoop-plugin/src/main/resources/dependencyPattern.json
```diff
@@ -5,7 +5,7 @@
       "message": "Open source versions(org.apache.*) of artifacts are not allowed, please use linkedin artifacts(com.linkedin.*). Open source versions will be completely blocked after March 1st 2016",
```
erwa: Can you also update this message with the specific dependencies?

erwa: Also, this is very LinkedIn-specific. Is it possible for this warning only to be applied internally? People using the LiHadoopPlugin outside of LinkedIn probably want to depend on org.apache artifacts or their own company-specific artifacts.

convexquad:
+1

Can you remove Tez from the blacklist? If we published LinkedIn-specific versions, we did so only with old versions of Tez - people would want to use newer versions.

I kind of like the current warning message, except that I think it should specifically mention org.apache.(hadoop, hive, oozie, pig) instead of org.apache.*.

li-hadoop-plugin/src/main/resources/dependencyPattern.json
nntnag17: The dependency plugin is actually disabled in the hadoop-plugin. We've enabled it only in li-hadoop-plugin, which is a Linkedin specific version of the plugin.

erwa: Good to know. In that case, I'm fine checking this in, after Alex's suggestions of
* Removing Tez from the list
* Explicitly list the Apache artifacts that are blacklisted rather than use a wildcard

are addressed.

nntnag17: I've addressed the comments and  just updated the PR. Now the complete error message looks like:

*WARN: Invalid dependency (org.apache.pig:pig:0.10.0), Open source versions of pig artifacts(org.apache.pig) are not allowed, please use linkedin artifacts(com.linkedin.pig). Open source versions will be completely blocked after March 1st 2016*

nntnag17:
Updated the PR after addressing the comments.

erwa:
LGTM

convexquad:
@nntnag17 Can you merge this?

nntnag17:
Merged

PR #31, opened 2015-11-12
Add generic GatewayCommand helper class for running remote commands t…

convexquad:
…hrough a gateway and make some style fixes

Nothing uses the new GatewayCommand helper class yet, so this is safe. I'm planning to set the Pig, Spark, and (upcoming) Hive Plugins to all use this helper class for their tasks that run scripts / jobs through the gateway.


convexquad:
This is pretty safe - nothing uses the GatewayCommands library code yet, so I'm going to +1 and merge it so I can rebase the rest of my work on the commit.

PR #32, opened 2015-11-14
Fix hadoop-plugin subproject unit tests

convexquad:


convexquad:
+1 Verified that "gradle test" passes

PR #33, opened 2015-11-14
Refactor Hadoop zip functionality into separate ZipPlugin class and e…

convexquad:
…stablish a clear patter in the HadoopPlugin class where cross-plugin task dendencies are setup

### DIFF
--- a/VERSIONS.md
+++ b/VERSIONS.md
@@ -17,8 +17,13 @@ the License.
 Note that the LinkedIn build system occasionally requires that we skip a version bump, so you will
 see a few skipped version numbers in the list below.
 
+0.7.3
+
+* Refactor Hadoop zip functionality into separate ZipPlugin class and establish a clear place in the HadoopPlugin class where task dependencies between plugins are set
+
 0.7.2
 
+* Refactor Hadoop zip functionality into separate ZipPlugin class and establish a clear place in the HadoopPlugin class where task dependencies between plugins are set
 * LIHADOOP-16016 Minor bug fix for li-hadoop-plugin
 * Fix hadoop-plugin subproject unit tests
 * Add generic GatewayCommand helper class for running remote commands through a gateway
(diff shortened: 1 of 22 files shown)


convexquad:
I was trying to update the li-photon-plugin to use the latest Hadoop Plugin internally, but it was failing in PCL due what was essentially a backwards compatibility problem.

The problem is that the Hadoop Plugin is becoming complicated enough that we can't make ordering assumptions about when its various sub-plugins are applied. For example, if the ScmPlugin declares task A, the SparkPlugin shouldn't assume that its task B dependsOn A, because that creates an ordering dependency that says the ScmPlugin must be applied before the SparkPlugin.

Such implicit ordering dependencies were fine when the Hadoop Plugin was small, but now that it's growing, that web of dependencies got too complicated somewhere along the 0.5.x or 0.6.x series.

In this PR I do a few main things:
1. Create the hadoopRuntime configuration right in the main HadoopPlugin class. All sub-plugins are safe to assume that this configuration always exists.
2. Pull out the Zip-related tasks (startHadoopZips, buildHadoopZips, the HadoopZipExtension class, etc.) into their own HadoopZipPlugin class and package. This looks right. The ScmPlugin is now just about metadata-ish and source code stuff, which also looks right.
3. Establish a central place in the HadoopPlugin class that sets cross-plugin task dependencies (for example, the output of the ScmPlugin tasks gets added to the Hadoop zips). This happens after all the sub-plugins have been applied.
4. Remove cross sub-plugin dependencies since they are established in (3) above.
5. Minor code and Javadoc cleanups

convexquad:
Since this is a large PR, I'm not really expecting anybody to review it. I'll leave it open for a couple days in case anyone wants to review or just add comments!

Even though I have been careful and added some new unit tests, I won't be too surprised if this breaks things. Since the 0.7.x series is still a development series (it's not an endorsed version on go/HadoopDSL), I think that's fine.

PR #35, opened 2015-11-23
Add Hadoop zip tasks directly instead of through project.afterEvaluate

convexquad:
Clean up hadoop-plugin unit tests and start to add tests in li-hadoop-plugin-test
Start removing ability to disable sub-plugins
Update Gradle to version 2.9 and enable the Gradle daemon - tests run much faster


convexquad:
In this Pull Request:

-Add Hadoop zip tasks directly instead of through project.afterEvaluate
This will generally improve the robustness of these tasks as they are now created as soon as people configure the hadoopZip extension in their build.gradle file. This also makes it possible to simplify the unit tests.

-Clean up hadoop-plugin unit tests and start to add tests in li-hadoop-plugin-test
I added a "li-hadoop-plugin-test" subproject which tests creating the CRT zip - we can start adding other LinkedIn-specific tests here

-Start removing ability to disable sub-plugins
Start getting rid of the "disableXXXPlugin" properties to turn off individual sub-plugins. These properties were added in case there was some catastrophic bug in a sub-plugin that would break user projects - then they could use the property to just disable the sub-plugin. Since the Hadoop Plugin is getting more mature, let's stop using this pattern and start removing it from the existing sub-plugins.

-Update Gradle to version 2.9 and enable the Gradle daemon - tests run much faster
The tests now run SEVERAL MINUTES faster - yes!

convexquad:
Merging - I took away the update to Gradle 2.9 as some LinkedIn Gradle Plugins don't work on 2.9 yet. However, I kept the Gradle daemon enabled - tests still run much faster.

Issue #42, opened 2015-12-22
ligradle azkabanUpload fails with "ClassNotFoundException: org.apache.http.client.methods.HttpPost"


erwa:
Internal JIRA: LIHADOOP-16591

This seems to be caused by the fact that the hadoop-plugin does not declare an explicit dependency on `org.apache.httpcomponents:httpclient`, which contains the `HttpPost` class.

erwa:
I believe the reason the Hadoop Plugin builds successfully without this explicit dependency is that Gradle itself pulls in the dependency (See https://github.com/gradle/gradle/blob/master/gradle/dependencies.gradle#L74).

PR #43, opened 2015-12-22
LIHADOOP-16591: ligradle azkabanUpload fails with "ClassNotFoundException: org.apache.http.client.methods.HttpPost"

erwa:
Fix for issue #42.


convexquad:
Thanks for the fix! I hadn't noticed this was happening. This jar must have been supplied automatically on the classpath from RUM, but then perhaps a new version of RUM stopped supplying it.

+1

PR #44, opened 2016-01-13
Update build system to publish to plugins.gradle.org

convexquad:
Additionally simplify the build system to just use a single standard build.gradle file, instead of the silly and confusing LinkedIn convention of having a defaultEnvironment.gradle file.


convexquad:
+1'ing myself

PR #45, opened 2016-01-13
Fixes for publishing to plugins.gradle.org

convexquad:
Fixes - this works. I just published the Hadoop Plugin on plugins.gradle.org.


convexquad:
+1'ing to merge the fix to be able to publish

PR #49, opened 2016-02-18
LIHADOOP-6730: Add 'clusterProvided' configuration for dependency man…

nntnag17:
…agement

reference: https://github.com/rdsr/blog/blob/master/2015-12-28.md

### DIFF
--- a/hadoop-plugin/src/main/groovy/com/linkedin/gradle/hadoop/HadoopPlugin.groovy
+++ b/hadoop-plugin/src/main/groovy/com/linkedin/gradle/hadoop/HadoopPlugin.groovy
@@ -29,6 +29,9 @@ import org.gradle.api.Project;
 import org.gradle.api.Task;
 import org.gradle.api.artifacts.Configuration;
 import org.gradle.api.plugins.JavaPlugin;
+import org.gradle.plugins.ide.eclipse.EclipsePlugin;
+import org.gradle.plugins.ide.idea.IdeaPlugin;
+import org.gradle.plugins.ide.idea.model.IdeaModule;
 
 /**
  * HadoopPlugin is the class that implements our Gradle Plugin.
@@ -42,7 +45,7 @@ class HadoopPlugin implements Plugin<Project> {
    */
   @Override
   void apply(Project project) {
-    addHadoopConfiguration(project);
+    addHadoopConfigurations(project);
     project.getPlugins().apply(getAzkabanPluginClass());
     project.getPlugins().apply(getDependencyPluginClass());
     project.getPlugins().apply(getHadoopDslPluginClass());
@@ -55,12 +58,71 @@ class HadoopPlugin implements Plugin<Project> {
   }
 
   /**
+   * Create hadoop configurations - hadoopRuntime and clusterProvided
+   * @param project The Gradle project
+   */
+  void addHadoopConfigurations(Project project) {
+    addHadoopRuntimeConfiguration(project);
+    addClusterProvidedConfiguration(project);
+  }
+
+  /**
+   * Prepare the "clusterProvided" Hadoop configuration for the project.
+   *
+   * @param project The Gradle project
+   * @return The clusterProvided configuration
+   */
+  Configuration addClusterProvidedConfiguration(Project project) {
+
+    // create clusterProvided configuration and set it as a private configuration.
+    Configuration clusterProvided = project.getConfigurations().create("clusterProvided");
+    clusterProvided.setVisible(false);
+
+    // we get default, testCompile, compile and runtime configurations from java plugin.
+    project.getPlugins().withType(JavaPlugin) {
+
+      // remove the clusterProvided dependencies from default configuration.
+      clusterProvided.dependencies.all {
+        dep -> project.getConfigurations().getByName('default').exclude group: dep.group, module: dep.name
+      }
+
+      // testCompile should extend from clusterProvided so that all the clusterProvided dependencies including transitive
+      // dependencies are on the classpath to facilitate testing.
+      project.getConfigurations().getByName("testCompile").extendsFrom(clusterProvided);
+
+      // add clusterProvided jars on the classpath, this will not add transitive dependencies of each jar.
+      project.sourceSets.main.compileClasspath += clusterProvided
+      project.javadoc.classpath += clusterProvided
+
+      // Idea doesn't understand the above classpath modification. Adding it explicitly. Java plugin
+      // is also required for getting non empty scopes.
+      project.getPlugins().withType(IdeaPlugin) {
+
+        IdeaModule ideaModule = project.idea.module;
+
+        // Additional checks for map keys.
+        if( ideaModule.scopes.containsKey("PROVIDED") && ideaModule.scopes.PROVIDED.containsKey("plus") ) {
+          project.idea.module.scopes.PROVIDED.plus += [clusterProvided]
+        }
+      }
+    }
+
+    // Eclipse understands the classpath, but add it, to be on the safe side.
+    project.getPlugins().withType(EclipsePlugin) {
+      project.eclipse.classpath.plusConfigurations += [clusterProvided]
+    }
+
+    return clusterProvided;
+  }
+
+
+  /**
    * Prepare the "hadoopRuntime" Hadoop configuration for the project.
    *
    * @param project The Gradle project
    * @return The hadoopRuntime configuration
    */
-  Configuration addHadoopConfiguration(Project project) {
+  Configuration addHadoopRuntimeConfiguration(Project project) {
     Configuration hadoopRuntime = project.getConfigurations().create("hadoopRuntime");
 
     // For Java projects, the Hadoop configuration should contain the runtime jars by default.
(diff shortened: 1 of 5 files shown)


erwa:
I suggest just naming the configuration "provided" rather than "clusterProvided" since "provided" is a term used in Maven and elsewhere.

nntnag17:
Addressed Anthony's comment

erwa:
Looks good otherwise

hadoop-plugin/src/main/groovy/com/linkedin/gradle/hadoop/HadoopPlugin.groovy
```diff
@@ -29,6 +29,8 @@ import org.gradle.api.Project;
 import org.gradle.api.Task;
 import org.gradle.api.artifacts.Configuration;
 import org.gradle.api.plugins.JavaPlugin;
+import org.gradle.plugins.ide.idea.IdeaPlugin;
+import org.gradle.plugins.ide.eclipse.EclipsePlugin;
```
convexquad: Could you re-order the idea / eclipse plugin imports alphabetically

hadoop-plugin/src/main/groovy/com/linkedin/gradle/hadoop/HadoopPlugin.groovy
```diff
@@ -55,12 +57,64 @@ class HadoopPlugin implements Plugin<Project> {
...
+      project.javadoc.classpath += provided
+
+      // Idea doesn't understand the above classpath modification. Adding it explicitly.
+      project.getPlugins().withType(IdeaPlugin) {
+        project.idea.module.scopes.PROVIDED.plus += [provided]
+      }
+    }
+
+    // Eclipse understands the classpath, but add it, to be on the safe side.
+    project.getPlugins().withType(EclipsePlugin) {
```
convexquad: Should this be inside the project.getPlugins().withType(JavaPlugin) { } code block?

convexquad:
I added a couple of specific comments. Hold off on merging this commit yet. I'm not sure if this is the right model for the provided configuration.

@nntnag17 can you look at the documentation for https://github.com/nebula-plugins/gradle-extra-configurations-plugin, and determine if:
1. It's possible for us to re-use their plugin directly, or if there's a good Hadoop-specific reason why we should not do so
2. Whether or not we've got the right model for our provided configuration. The Nebula plugin documentation mentions their model is the Maven provided configuration, which says:

provided - this is much like compile, but indicates you expect the JDK or a container to provide it at runtime. It is only available on the compilation and test classpath, and is not transitive

This seems like the model that we want, except that:
1. I think (but I'm not sure) that what we want is that the provided jars are available and transitive in the compile, testCompile, runtime and testRuntime configurations; but are not available and not transitive in the hadoopRuntime configuration.
2. Of course for us, the grid will provide the dependency at hadoopRuntime (instead of the JDK like in the Maven statement)

But perhaps this isn't the right model. @erwa we might also want to get Carl's opinion on this.

convexquad:
I also think we should probably call it clusterProvided (or gridProvided or hadoopProvided?) to avoid potential name clashes with any other plugins that declare a "provided" configuration, especially if our model ends up being slightly different than the Maven model.

nntnag17:
@convexquad I think the only reason why we may not want to use the mentioned nebula plugin is that it assumes that the java plugin is already applied. Hadoop plugin is also applied to projects which do not apply java plugin (Hive, Pig project), in such projects the nebula plugin throws Runtime exception. The reason being when java plugin is not applied, project.idea.module.scopes is an empty Map and nebula plugin tries to call project.idea.module.scopes.PROVIDED. Although there might be ways to make it work but I'm not sure if it is worth the effort. 

Also, the model that you proposed seems pretty good to me. Only modification in the above code to fit the model will be to enable transitive dependencies in compile configuration as well. Please let me know your thoughts and I'll update the PR.

erwa:
I personally think the "correct" model is for compile to extend provided and then for default to exclude all provided dependencies. This is exactly what is described in http://www.sinking.in/blog/provided-scope-in-gradle/

However, as we've discovered, Scala sbt and the Grape resolver do not honor the `<exclude>`s

nntnag17:
Hey @convexquad, 

Have we decided on something yet on what approach should we follow?

convexquad:
Sorry @nntnag17, I have just been busy with my project. I still need to take another look at the Nebula plugin and to consider your comments. I'll try to get back to you early next week.

convexquad:
Just wanted to let you know that I am looking at this - I haven't forgotten it.

hadoop-plugin/src/test/groovy/com/linkedin/gradle/hadoop/HadoopPluginTest.groovy
```diff
@@ -43,7 +43,9 @@ class HadoopPluginTest {
   public void testApplyHadoopPlugin() {
     plugin.apply(project);
 
-    // Check that the hadoopRuntime configuration was created successfully.
+    // Check that hadoop configurations were created successfully.
```
convexquad: Can you find a way to add a unit test that adds a provided dependency, then checks that the provided dependency is NOT in the compile, runtime, default and hadoopRuntime configurations, but IS in the testCompile configuration?

convexquad:
Ok, I read http://www.sinking.in/blog/provided-scope-in-gradle - thanks @erwa for posting that link. That page basically explains the range of options, and explains some of the code I see in https://github.com/nebula-plugins/gradle-extra-configurations-plugin/blob/master/src/main/groovy/nebula/plugin/extraconfigurations/ProvidedBasePlugin.groovy.

@nntnag17 can you update the PR for the comments I made?

I still think we should change the name of the configuration to the "clusterProvided" (or "gridProvided") configuration to distinguish the intention between the Maven-style "provided" configuration and our more Hadoop-focused provided configuration - for example, the li-hadoop-plugin subproject could include a LinkedIn-specific list of dependencies that are known to exist on our clusters and throw an error if you put them in a configuration other than clusterProvided.

Additionally, we're going to need to double-check the Ivy output files that get written when we build artifacts that use this new provided configuration. I think in Anant's current design, they will be fine, since we have testCompile extending provided, so that in the Ivy file the provided dependencies will show up as transitive dependencies, but only for the testCompile configuration (which I think is ok).

convexquad:
@erwa Do we ever need sbt or the Grape resolver at LinkedIn? If those tools are broken, I don't think it's our responsibility for the Hadoop Plugin to work around their problems. I do think I remember that Grape is the resolver used for Dali - is that right? In that case, then yes, we should work around the problem as we don't want any artifacts published by the Hadoop Plugin to break Dali.

erwa:
@convexquad: I agree with you that the configuration name should probably be `clusterProvided` or `gridProvided` to distinguish from Maven `provided` behavior.

Hive, Pig, and Dali all use Grape for doing dependency resolution, so they're all affected by this.

convexquad:
@nntnag17 Let us know when you've got an updated PR ready!

hadoop-plugin/src/main/groovy/com/linkedin/gradle/hadoop/HadoopPlugin.groovy
```diff
@@ -55,12 +57,64 @@ class HadoopPlugin implements Plugin<Project> {
...
+      // testCompile should extend from provided so that all the provided dependencies including transitive
+      // dependencies are on the classpath to facilitate testing.
+      project.getConfigurations().getByName("testCompile").extendsFrom(provided);
+
+      // add provided jars on the classpath, this will not add transitive dependencies of each jar.
+      project.sourceSets.main.compileClasspath += provided
+      project.javadoc.classpath += provided
+
+      // Idea doesn't understand the above classpath modification. Adding it explicitly.
+      project.getPlugins().withType(IdeaPlugin) {
```
nntnag17: The IdeaPlugin maintains a map of configurations called scopes. When JavaPlugin is applied then the scopes map has a key called PROVIDED in which we add the clusterProvided configuration dependencies. I think it makes sense to keep it inside the block which checks for JavaPlugin since without the JavaPlugin, we have nothing to add to IdeaPlugin.

nntnag17:
Hey @convexquad, I've updated the PR. Sorry for the delay.

convexquad:
I committed a PR that bumps to 0.8.0 already - you'll need to go to 0.8.1. After you do this, you are +1 to merge. Thanks for the amazing contribution! I'll publish a new version of li-hadoop-plugin after your merge.

Also, thanks for the great unit tests!!!

After the new version is published, do you have time to publish an RB for the Hadoop Starter Kit that alters some of the dependencies to use the new provided configuration? I'm eager to see what it would look like, and we need to double check that resulting .ivy files.

convexquad:
Additionally, we can think about some kind of li-hadoop-plugin subproject subclass that checks your compile and runtime dependencies against known dependencies that area already on our Hadoop grids, and outputs a warning telling you that you should declare them as clusterProvided dependencies. Maybe this is something we can plan for with Suja and Shankar.

nntnag17:
Hey @convexquad, I've merged the PR. I'll send an RB to for HSK soon. 

Also I think the idea of li-hadoop-plugin checking for the dependencies already present in the GRID is a very good idea. I think Carl mentioned something related to discovery service which can provide this information to us. I'll also talk to Shankar regarding this.

PR #51, opened 2016-03-16
Improved hadoopClosure and evalHadoopClosure Hadoop DSL methods that …

convexquad:
…are way more expressive


convexquad:
Merging my new features!

PR #52, opened 2016-03-17
Gradle version bumped to 2.10

convexquad:
Just bumping the Gradle version from 2.7 -> 2.10


convexquad:
Merging - this is safe to merge

PR #54, opened 2016-03-18
Various minor improvements for the Hadoop DSL from my backlog

convexquad:
-Remove unused lookupRef method and add lookup / lookupRef tests
-Remove unnecessary mavenCentral repository declaration from tests
-Print the stacktrace when tests from hadoop-plugin-test fail that should have passed
-Add the more explicit "set hadoopProperties" as a synonym for "set confProperties"
-Worklow targets method and Job depends method can take explicit lists
-Remove final declaration from non-final properties in a couple of job types
-Minor alphabetization and code cleanups


convexquad:
+1 Merging my PR!

PR #61, opened 2016-05-12
Large number of minor fixes to reduce LinkedIn codestyle and codenarc…

convexquad:
… check failures


convexquad:
After this PR, there are far fewer checkstyle and codenarc check failures in the output HTML reports. There are still a bunch of them left, but for some of them we'll probably configure the checkstyle rules to support some of the idioms we are using.

convexquad:
I'm going to publish another PR first addressing the build stability for LinkedIn builds

PR #62, opened 2016-05-13
Increase build robustness for LinkedIn ligradle builds, which keep ch…

convexquad:
…anging how SNAPSHOT versions work


convexquad:
Merging!

PR #77, opened 2016-08-05
Added support for submitting python Spark applications with Spark job type. Corresponsing tests are added.

coderant:
some minor import optimization in OozieDslCompiler
Thanks for the help!
Documentation needs to be updated for Spark job type.


convexquad:
+1 Thanks for the second contribution!

PR #79, opened 2016-08-09
Do not include any folder named 'resources' in the embedded sources zip

convexquad:


convexquad:
I have decided against this PR. My reason is that excluding resources from the embedded sources zip could prevent the Hadoop Team from being able to diagnose the cause of a job failure if somehow the failure is related to files inside the resources directories.

People that want to exclude the resources folder from the sources zip can simply use the writeScmPluginJson task and then edit the .scmPlugin.json file to add an entry for the resources folder.

PR #82, opened 2016-08-12
add python spark support for runSparkJob

coderant:
hard remote reset saved me again... 
must not get too comfortable with it...


convexquad:
Thanks again!!!

convexquad:
Ah, can you bump the version in gradle.properties? Then I'll merge it

PR #88, opened 2016-08-18
Add additional Hadoop DSL workflow flowDepends override that matches …

convexquad:
…the override for the targets method


convexquad:
+1 Enhancement requested by a user

Issue #84, opened 2016-08-12
Additional enhancements for azkabanUpload task


convexquad:
We have been working on improving the azkabanUpload task. Thanks to @pranayhasan for his great work so far!

In the commit https://github.com/linkedin/linkedin-gradle-plugin-for-apache-hadoop/commit/9a27f6f073292c9b0b7963a569c4e993fd29fd9f I noted that I would love to get the following further enhancements:

1. ~~Perfecting console printing of the project values and query questions~~
2. ~~Displaying the current values within the query questions~~
3. ~~The double slash (e.g. ":8443//manager") in the Azkaban project URL that prevents you from going directly to the project page~~
4. Also the enhancement to show the Azkaban zips you have configured in your hadoopZip block

These would all serve to further improve the azkabanUpload task. Please go to the conversation for that commit for further details about these enhancements.

convexquad:
Never mind on the double-slash in the manager URL, it's fixed

convexquad:
I knocked out "2. Displaying the current values within the query questions" in my PR today, I'll cross it out

convexquad:
I have crossed out number 1, which is fixed in https://github.com/linkedin/linkedin-gradle-plugin-for-apache-hadoop/pull/99. @pranayhasan I would still love to get enhancement number 4 on printing the zips the user has declared in their hadoopZip block!

PR #100, opened 2016-09-11
Display configured Hadoop zips during azkabanUpload task

pranayhasan:
Also replaced magic with holdem and canasta with war, in javadocs.

### DIFF
--- a/hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/AzkabanHelper.groovy
+++ b/hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/AzkabanHelper.groovy
@@ -16,10 +16,12 @@
 package com.linkedin.gradle.azkaban;
 
 import com.linkedin.gradle.util.HtmlUtil;
-
+import com.linkedin.gradle.zip.HadoopZipExtension;
 import org.apache.http.impl.client.CloseableHttpClient;
 import org.apache.http.impl.client.HttpClientBuilder;
 import org.apache.http.util.EntityUtils;
+import org.gradle.api.Project;
+import org.gradle.api.file.CopySpec;
 
 import java.security.cert.CertificateException;
 import java.security.cert.X509Certificate;
@@ -229,9 +231,10 @@ class AzkabanHelper {
    * Configures the fields for azkabanUploadTask.
    *
    * @param azkProject The AzkabanProject
+   * @param project The Gradle Project
    * @return boolean to save the entered fields to .azkabanPlugin.json
    */
-  static boolean configureTask(AzkabanProject azkProject) {
+  static boolean configureTask(AzkabanProject azkProject, Project project) {
     def console = System.console();
     if (console == null) {
       String msg = "\nCannot access the system console. To use this task, explicitly set JAVA_HOME to the version specified in product-spec.json (at LinkedIn) and pass --no-daemon in your command.";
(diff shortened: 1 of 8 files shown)


hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/AzkabanHelper.groovy
```diff
@@ -267,7 +270,23 @@ class AzkabanHelper {
...
+          def counter = 1;
+          zipMap.each { String zipName, CopySpec copySpec ->
+            logger.lifecycle("${counter++}. ${zipName.replace("hadoopZip","").replace("azkaban","")} - Enter \"${zipName}HadoopZip\" to use this zip");
+          }
+          logger.lifecycle("(Or you can use the name of any other Gradle zip task)");
+        }
+
+        //waiting for the above logger statements to flush
+        sleep(500);
+        input = console.readLine(" > New Azkaban Zip task [enter to accept '${azkProject.azkabanZipTask}']: ");
```
pranayhasan: Again there is an interruption from build message. It would be nice to know if there is a way to delay the interruption from build. One other way to make everything look even, is to take input by appending with the build message itself, as you did in your previous commit using
`console.format("<message>").flush();`

hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/AzkabanHelper.groovy
```diff
@@ -267,7 +270,23 @@ class AzkabanHelper {
...
+        HadoopZipExtension hadoopZipExtension = project.extensions.getByName("hadoopZip");
+        Map<String, CopySpec> zipMap = hadoopZipExtension.getZipMap();
+
+        if (zipMap.isEmpty()) {
+          logger.lifecycle("No zips configured in the hadoopZip block");
+        } else {
+          logger.lifecycle("\nFound the following zips in the hadoopZip block:");
+          def counter = 1;
+          zipMap.each { String zipName, CopySpec copySpec ->
+            logger.lifecycle("${counter++}. ${zipName.replace("hadoopZip","").replace("azkaban","")} - Enter \"${zipName}HadoopZip\" to use this zip");
```
pranayhasan: replace("azkaban") seems to be linkedin specific to me now. Do you think we have to remove it?

hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/AzkabanHelper.groovy
```diff
@@ -229,9 +231,10 @@ class AzkabanHelper {
    * Configures the fields for azkabanUploadTask.
    *
    * @param azkProject The AzkabanProject
+   * @param project
```
convexquad: @param project The Gradle project

hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/AzkabanHelper.groovy
convexquad: I actually don't think you need to replace either hadoopZip or azkaban out of zipName - I think we should just show the exact name the user used in the hadoopZip block, e.g.

hadoopZip {
  zip("myAzkabanZip") {
    ...
  }
}

Then when they run the task, they should see something like:

1. myAzkabanZip - Enter "myAzkabanZipHadoopZip" to use this zip

pranayhasan: Ok this looks good, as it also makes it generic.

hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/AzkabanHelper.groovy
pranayhasan: Addressed it

hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/AzkabanPlugin.groovy
```diff
@@ -232,6 +232,7 @@ class AzkabanPlugin implements Plugin<Project> {
   /**
    * Load the Azkaban project properties defined in the .azkabanPlugin.json file.
    *
+   * @param project
```
convexquad: param project The Gradle project

convexquad:
@pranayhasan This looks good! I added one more little comment, after your resolve it I will merge this commit and close the associated GitHub issue

convexquad:
+1

convexquad:
Closes #84

Issue #102, opened 2016-09-12
The azkabanUpload task should automatically go into edit mode if there are missing required fields


convexquad:
Currently the azkabanUpload task will ask you if you want to edit fields, even if some of them are blank:
```
Entering interactive mode. You can use the -PskipInteractive command line parameter to skip interactive mode and ONLY read from the .azkabanPlugin.json file.

Azkaban Project Name: 
Azkaban URL: https://ltx1-holdemaz01.grid.linkedin.com:8443
Azkaban User Name: abain
Azkaban Zip Task: 
> Building 93% > :xgboost-demo:azkabanUpload > Want to change any of the above? [y/N]: 
```
If the user is missing any of the required fields, we should just go directly into edit mode.

PR #103, opened 2016-09-12
The azkabanUpload task should automatically go into edit mode if ther…

pranayhasan:
…e are missing required fields


convexquad:
+1

convexquad:
Closes #102

PR #107, opened 2016-09-16
Azkaban CLI features

pranayhasan:
Create Project, Flow Status, Execute Flow and Cancel Flow

### DIFF
--- a/hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/AzkabanCreateProjectTask.groovy
+++ b/hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/AzkabanCreateProjectTask.groovy
@@ -1,0 +1,94 @@
+/*
+ * Copyright 2015 LinkedIn Corp.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.linkedin.gradle.azkaban;
+
+import com.linkedin.gradle.util.HttpUtil;
+
+import org.apache.http.client.utils.URIBuilder;
+
+import org.gradle.api.DefaultTask;
+import org.gradle.api.GradleException;
+import org.gradle.api.tasks.TaskAction;
+
+import org.json.JSONObject;
+
+/**
+ * AzkabanCreateProjectTask creates a new project in Azkaban.
+ */
+class AzkabanCreateProjectTask extends DefaultTask {
+
+  AzkabanProject azkProject;
+
+  /**
+   * The Gradle task action for creating project in Azkaban.
+   */
+  @TaskAction
+  void create() {
+    if (azkProject.azkabanUrl == null) {
+      throw new GradleException("Please set azkaban.url in the .azkabanPlugin.json file in your project directory.");
+    }
+    createAzkabanProject(AzkabanHelper.readSession());
+  }
+
+  /**
+   * Creates a project in Azkaban from project name in .azkabanPlugin.json file.
+   *
+   * @param sessionId The Azkaban session id. If this is null, an attempt will be made to login to Azkaban.
+   */
+  void createAzkabanProject(String sessionId) {
+    try {
+      sessionId = AzkabanHelper.resumeOrGetSession(sessionId, azkProject);
+
+      def console = System.console();
+      if (console == null) {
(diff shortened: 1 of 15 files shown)


hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/AzkabanCreateProjectTask.groovy
```diff
@@ -0,0 +1,89 @@
...
+      logger.lifecycle("Enter Project description:");
+      String description = System.console().readLine(" Input>");
+      while (description.isEmpty()) {
+        logger.lifecycle("Enter a non-empty Project description");
+        description = System.console().readLine(" Input>");
+      }
+
+      URI createURI = new URIBuilder(azkProject.azkabanUrl)
+          .setPath("/manager")
+          .setParameter("session.id", sessionId)
```
nntnag17: The values "session.id", "action", "name" etc are duplicated at various places. Can we separate them out to an Enum or a class and use that instead. Something like AzkabanParams.SESSION_ID or AzkabanParams.ACTION.

hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/AzkabanCreateProjectTask.groovy
```diff
@@ -0,0 +1,89 @@
...
+          .setParameter("action", "create")
+          .setParameter("name", azkProject.azkabanProjName)
+          .setParameter("description", description.trim().toString())
+          .build();
+
+      String response = AzkabanHelper.responseFromPOST(createURI);
+
+      if (response.toLowerCase().contains("error")) {
```
nntnag17: Similar to above comment, can we have an AzkabanResponses enum or class for responses such as error etc?

hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/AzkabanExecuteFlowTask.groovy
```diff
@@ -0,0 +1,162 @@
...
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.linkedin.gradle.azkaban;
+
+import org.apache.http.client.utils.URIBuilder;
+
+import org.gradle.api.DefaultTask;
+import org.gradle.api.GradleException
```
nntnag17: Missing semicolons

hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/AzkabanExecuteFlowTask.groovy
```diff
@@ -0,0 +1,162 @@
...
+
+      List<String> flows = AzkabanHelper.fetchSortedFlows(new JSONObject(response), azkProject);
+
+      Set<String> indexSet;
+
+      if (AzkabanPlugin.interactive) {
+        logger.lifecycle("-----    ----------");
+        logger.lifecycle("INDEX    FLOWS LIST");
+        logger.lifecycle("-----    ----------");
+        for(int i=0; i<flows.size(); i++) {
```
nntnag17: flows.each?

nntnag17: or if you need index eachWithIndex?

pranayhasan:
Want to know if there are any quick additions/IO changes. Would cleanup once the view is finalized.

convexquad:
@pranayhasan Since this is a larger RB, give me a few more days to work on the review

convexquad:
@pranayhasan @nntnag17 I want to ask that you make a major design change with this PR. I don't want to repeat the same mistake as with the azkaban-cli (internal to LinkedIn) multiproduct, where it is hard to re-use the code because it assumes that you are interacting with it in a specific way.

I would like to ask you to create a new Gradle subproject for interacting with Azkaban that just contains a series of library methods with high-level API's, e.g. createAzkabanProject(AzkabanProject azkabanProject, String sessionId) that don't assume anything about Gradle or the Hadoop DSL / Plugin.

Then leave all the Gradle tasks in place in the hadoop-plugin project. Probably there will still be the same classes that there are now (AzkabanCreateProjectTask, etc.), however instead of having the code in them that creates the project, they should call the API from the other subproject.

convexquad:
Second comment - can you include as many unit tests as you can for any standalone helper functions (i.e. the functions that don't make remote calls to Azkaban). For the functions that actually call Azkaban, testing will be more difficult. We should develop tests for these methods, but I think we can do it as a separate PR.

pranayhasan:
Thanks @convexquad for the recommendations. I agree with the concerns raised by you, as I haven't been able to reuse the Azkaban-CLI though it is inherently calling the same API. Would work on refactoring to make the API calls independent to that of hadoop-plugin. I will continue to commit on the existing PR and ping you once I'm done.

convexquad:
@pranayhasan Do you have an ETA on the refactoring work for this Pull Request? It would be an awesome enhancement for the Hadoop Plugin

pranayhasan:
@convexquad I had other prioritized work, so had to put this on backlog. I started refactoring it. You can expect updated PR by early next week.

pranayhasan:
Opened a new PR discarding this.

PR #112, opened 2016-10-05
Enable properly redeclaring the paths your Hadoop DSL job reads and w…

convexquad:
…rites


convexquad:
Verified that this fixes my extraneous Hadoop DSL static checker messages

PR #114, opened 2016-10-12
Display custom message after azkabanUpload task

pranayhasan:
This can be useful to notify users on recent developments after azkabanUpload task completes, especially specific to Linkedin. As the file with the message is maintained in an internal repository, the messages are visible only to company users, and no code changes are needed to make changes to the message. Additionally, empty file displays no message to users.

### DIFF
--- a/hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/AzkabanPlugin.groovy
+++ b/hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/AzkabanPlugin.groovy
@@ -38,7 +38,7 @@ import static com.linkedin.gradle.azkaban.AzkabanConstants.*;
 class AzkabanPlugin implements Plugin<Project> {
 
   boolean interactive = true;
-  private final static Logger logger = Logging.getLogger(AzkabanPlugin);
+  protected final Logger logger = Logging.getLogger(AzkabanPlugin);
 
   /**
    * Applies the AzkabanPlugin. This adds the Gradle task that builds the Hadoop DSL for Azkaban.
@@ -129,6 +129,7 @@ class AzkabanPlugin implements Plugin<Project> {
       doLast {
         if (interactive) {
           logger.lifecycle("\nUse -PskipInteractive command line parameter to skip interactive mode and ONLY read from the .azkabanPlugin.json file.");
+          printUploadMessage();
         }
       }
     }
@@ -197,6 +198,12 @@ class AzkabanPlugin implements Plugin<Project> {
    */
   AzkabanProject makeDefaultAzkabanProject(Project project) {
     return makeAzkabanProject();
+  }
+
+  /**
+   * Prints a message after upload to Azkaban is done. Subclasses can override this method.
+   */
+  void printUploadMessage() {
   }
 
   /**
--- a/li-hadoop-plugin/src/main/groovy/com/linkedin/gradle/liazkaban/LiAzkabanPlugin.groovy
+++ b/li-hadoop-plugin/src/main/groovy/com/linkedin/gradle/liazkaban/LiAzkabanPlugin.groovy
@@ -15,16 +15,19 @@
  */
 package com.linkedin.gradle.liazkaban;
 
-
 import com.linkedin.gradle.azkaban.AzkabanDslCompiler;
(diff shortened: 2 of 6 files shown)


hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/AzkabanPlugin.groovy
```diff
@@ -200,6 +201,12 @@ class AzkabanPlugin implements Plugin<Project> {
   }
 
   /**
+   * Prints a message after upload to Azkaban is done
```
convexquad: Can you add a note in the Javadoc comment that this method is intended to be overwritten by subclasses

li-hadoop-plugin/src/main/groovy/com/linkedin/gradle/liazkaban/LiAzkabanPlugin.groovy
```diff
@@ -56,4 +60,25 @@ class LiAzkabanPlugin extends AzkabanPlugin {
...
+   * upcoming changes which affects Azkaban users. Subclasses can override this method.
+   */
+  @Override
+  void printUploadMessage() {
+    try {
+      String message = MESSAGE_URL.toURL().getText();
+      if (!message.isEmpty() && message.length() > 0) {
+        logger.lifecycle("\n"+message);
+      }
+    } catch (HttpException ex) {
```
convexquad: Perhaps we can just catch Exception here. Then we can remove the dependency on HttpException in this class. Also, I don't think we should print the stacktrace here - that could be annoying for users. Instead, you might print e.getMessage() as a part of the logger.error message.

li-hadoop-plugin/src/main/groovy/com/linkedin/gradle/liazkaban/LiAzkabanPlugin.groovy
```diff
@@ -56,4 +60,25 @@ class LiAzkabanPlugin extends AzkabanPlugin {
...
+      logger.info("Failed to fetch message");
```
convexquad: Maybe logger.error instead of logger.info?

li-hadoop-plugin/src/main/groovy/com/linkedin/gradle/liazkaban/LiAzkabanPlugin.groovy
```diff
@@ -56,4 +60,25 @@ class LiAzkabanPlugin extends AzkabanPlugin {
...
+      ex.printStackTrace();
+    } catch (IOException ex) {
+      logger.info("Failed to fetch message");
+      ex.printStackTrace();
+    }
+
```
convexquad: Kill extra whitespace line

convexquad:
Be sure to check the Travis CI log - something is causing one of the tests to fail

pranayhasan:
Thanks Alex for the quick review!

hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/AzkabanPlugin.groovy
pranayhasan: Done

li-hadoop-plugin/src/main/groovy/com/linkedin/gradle/liazkaban/LiAzkabanPlugin.groovy
pranayhasan: Done

li-hadoop-plugin/src/main/groovy/com/linkedin/gradle/liazkaban/LiAzkabanPlugin.groovy
pranayhasan: Addressed

li-hadoop-plugin/src/main/groovy/com/linkedin/gradle/liazkaban/LiAzkabanPlugin.groovy
pranayhasan: Yes, this made Travis fail.

PR #119, opened 2016-10-27
LIHADOOP-24299 Add support in the Hadoop Plugin for choosing the defa…

convexquad:
…ult artifact for Hadoop CRT deployments

### DIFF
--- a/VERSIONS.md
+++ b/VERSIONS.md
@@ -17,6 +17,10 @@ the License.
 Note that the LinkedIn build system occasionally requires that we skip a
 version bump, so you will see a few skipped version numbers in the list below.
 
+0.11.0
+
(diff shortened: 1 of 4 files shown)


convexquad:
Testing: I made a SNAPSHOT build of the li-hadoop-plugin multiproduct that wraps this GitHub repository, and then made a SNAPSHOT build of ligradle-hadoop with ligradle-core 1.x.

I had println's in the code and verified that it hit the case in my code where it sets the magic "defaultArtifact" artifact classifier and that the artifact-spec.json selected this artifact as the default artifact.

Then for ligradle-hadoop I bumped ligradle-core to 2.x, rebuilt my SNAPSHOT, and verified that now the println told me it set the generateModuleArtifactSpec.defaultArtifact property, and I verified that the artifact-spec.json indeed selected the specified task as the default artifact. That should pretty much cover it.

ethankhall:
LGTM

PR #123, opened 2016-11-16
Minor build setup cleanups and improvements - Apply eclipse and idea …

convexquad:
…plugins to all projects

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,20 @@ The Plugin includes the **LinkedIn Gradle DSL for Apache Hadoop** (which we shal
 the "Hadoop DSL" for brevity), a language for specifying jobs and workflows for Hadoop workflow
 schedulers like Azkaban and Apache Oozie.
 
+#### Can I Benefit from the Hadoop Plugin and Hadoop DSL?
+
+The Hadoop Plugin is only useful if you are using Gradle as your build system. If you are using
+Azkaban, you should start using the Hadoop Plugin immediately and you should start using the Hadoop
+DSL as the primary mechanism by which you develop your Azkaban workflows. Read the Hadoop Plugin
+User Guide and the Hadoop DSL Language Reference to get started today.
+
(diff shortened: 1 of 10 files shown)


pranayhasan:
Don't include jline in azkaban-client build script as it overrides older version of jline included in hadoop-plugin. It caused hadoop validator to not work previously.

convexquad:
Whoops, I had stashed these changes from an earlier point in the commit history and forgot to rebase. I will remove jline from azkaban-client.

pranayhasan:
Oh, sorry. You have already taken care of it. I would want to test if Hadoop Validator works appropriately before pushing this version.

convexquad:
What do you use for testing the Hadoop Validator tasks? Can I just run the hadoopValidate task in the Hadoop Starter Kit?

pranayhasan:
Yes, you can. Follow go/hadoopvalidator to set hadoop-starter-kit to be able to run it. I'm really excited to use this enhancement soon. I would use it once you push this. 💯

pranayhasan:
Never mind, as you removed jline dependency for azkaban-client it would anyway work.

convexquad:
In the hadoop-plugin project I moved jline:jline:1.0 to a runtime dependency, since it isn't actually used for compilation (but is indeed necessary at runtime).

I tested all the tasks with the hello-pig-project in the HSK and they all work correctly. The pigDataExists task fails since one of the datasets used in the HSK is no longer available (the old MEMBER2 database).

I am going to merge this PR since my SNAPSHOT build works correctly with the HSK. This PR will help me complete my work on the Hadoop DSL Intellij syntax completion.

@pranayhasan The new pigValidate tasks are totally awesome! I am super impressed with them. It is almost Thanksgiving here in the US, but after I come back I will help promote these new tasks, and also I want to do a bit more refactoring with the code base so that we can keep the tasks in good shape moving forward.

PR #129, opened 2016-12-15
Add possibility to group jobs into subflows in Azkaban UI

rride:
Currently, Hadoop DSL doesn't support embedded flows (see http://azkaban.github.io/azkaban/docs/latest/#creating-flows).

This change allows to specify "groupJobs true" in the workflow DSL to convert it into an embedded Azkaban flow automatically.

### DIFF
--- a/hadoop-plugin-test/src/main/gradle/positive/groupingWorkflows.gradle
+++ b/hadoop-plugin-test/src/main/gradle/positive/groupingWorkflows.gradle
@@ -1,0 +1,59 @@
+buildscript {
+  dependencies {
+    classpath files("${project.pluginTestDir}/hadoop-plugin-${project.version}.jar", "${project.pluginTestDir}/hadoop-plugin-${project.version}-SNAPSHOT.jar")
+  }
+}
+
+apply plugin: com.linkedin.gradle.hadoop.HadoopPlugin
+
+// test grouping subflow jobs
+hadoop {
+  buildPath "jobs"
+  cleanPath false
+
+  workflow('groupingFlow') {
+    // job1 is not defined yet
+
+    groupJobs true
+
+    targets 'job1', 'groupingSubFlow'
+
+    noOpJob('job1') {
+      depends 'job2'
+    }
+
+    workflow('groupingSubFlow') {
+      groupJobs true
+
+      noOpJob('sub-job1') {
+      }
+
+      flowDepends 'job1'
+      targets 'sub-job1'
+    }
+
+    noOpJob('job2') {
+    }
+  }
+
+  workflow('nonGroupingFlow') {
+    targets 'job1', 'groupingSubFlow2'
+
+    noOpJob('job1') {
+      depends 'job2'
+    }
+
+    workflow('groupingSubFlow2') {
+      groupJobs true
+
+      noOpJob('sub-job1') {
+      }
+
+      flowDepends 'job1'
+      targets 'sub-job1'
+    }
+
+    noOpJob('job2') {
+    }
+  }
+}
--- a/hadoop-plugin/src/main/groovy/com/linkedin/gradle/hadoopdsl/job/SubFlowJob.groovy
+++ b/hadoop-plugin/src/main/groovy/com/linkedin/gradle/hadoopdsl/job/SubFlowJob.groovy
@@ -1,0 +1,98 @@
+/*
+ * Copyright 2015 LinkedIn Corp.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.linkedin.gradle.hadoopdsl.job;
+
+import com.linkedin.gradle.hadoopdsl.NamedScope;
+
+
+/**
+ * SubFlowJob is a job that groups all its children jobs into a parent node in Azkaban.
+ */
+class SubFlowJob extends StartJob {
+  /**
+   * Root of a jobs' graph that defines the group
+   */
+  String targetJobName;
+
+  /**
+   * Constructor for a SubFlowJob.
+   *
+   * @param jobName The job name
+   */
+  SubFlowJob(String jobName) {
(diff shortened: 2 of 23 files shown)


convexquad:
@rride Thanks for waiting - I will try to review this soon. I can't remember exactly why I did not compile Hadoop DSL subflows as Azkaban embedded flows in the first place - however, I *think* it is because you cannot have Azkaban embedded flows inside of embedded flows (it won't allow for recursive embedded flows). Although I am not 100% sure about that - there could have been another reason.

convexquad:
@rride Thanks for waiting, I have been leading a really important project for the Relevance Org in 2017, but I just finished a milestone for it and I have some time to work on this.

I remember the problem now. The issue is that with Azkaban subflows, the subflows show up on the project page. For example, lets say I have flow1 that depends on an embedded flow called subflow1. Then in the Azkaban UI for the project page, I'll see both flow1 and subflow1 listed on the screen.

Do you know any way to get rid of that? We have projects with tons and tons of subflows. All of the individual subflows will show up as executable flows on the Azkaban project page. If we could get rid of that, I would really like this enhancement.

rride:
Hi,

I don't think it is a blocker for merging the changes into the master because the behavior will not change unless you specify "groupJobs true". That means that the change doesn't impact any existing workflows.

We can make this to be the default behavior only if we receive a satisfactory reply from the Azkaban team.

convexquad:
@rride Since we didn't hear back from the Azkaban team, I am guessing there is not a way to disable the subflow from appearing on the Azkaban UI as its own flow. I know the Azkaban codebase ok, so I am going to go through the code and double check that this is indeed the case.

I want to be sure about this, because it would be nice to enable this feature by default. However, given that there is probably not a way in Azkaban to prevent the subflow from showing up in the UI as its own flow, I think your approach of making this optional (i.e. "groupJobs true") definitely makes sense.

Given that, let me do two things:
1. I'll go through Azkaban codebase as I mentioned and verify there is no way to disable the subflow from showing up on the UI
2. I'll do the code review for your Pull Request - probably I will ask you for a couple of updates

hadoop-plugin/src/main/groovy/com/linkedin/gradle/hadoopdsl/job/SubFlowJob.groovy
```diff
@@ -0,0 +1,100 @@
...
+  /**
+   * Helper method to set the properties on a cloned job.
+   *
+   * @param cloneJob The job being cloned
+   * @return The cloned job
+   */
+  SubFlowJob clone(SubFlowJob cloneJob) {
+    return ((SubFlowJob)super.clone(cloneJob));
+  }
+
```
convexquad: Kill extra whitespace line

hadoop-plugin/src/main/groovy/com/linkedin/gradle/hadoopdsl/job/SubFlowJob.groovy
```diff
@@ -0,0 +1,100 @@
...
+   * @param parentScope The parent scope in which to lookup the base properties
+   * @return The job properties map that holds all the properties that will go into the built job file
+   */
+  @Override
+  Map<String, String> buildProperties(NamedScope parentScope) {
+    Map<String, String> allProperties = super.buildProperties(parentScope);
+
+    // flow.name defines a root of jobs' dependency graph that will be used to group them
+    allProperties['flow.name'] = buildFileName(parentScope, targetJobName);
+
```
convexquad: Kill extra whitespace line

hadoop-plugin/src/main/groovy/com/linkedin/gradle/hadoopdsl/job/SubFlowJob.groovy
```diff
@@ -0,0 +1,100 @@
...
+package com.linkedin.gradle.hadoopdsl.job;
+
+import com.linkedin.gradle.hadoopdsl.NamedScope;
+
+
+/**
+ * SubFlowJob is a job that groups all its children jobs into a parent node in Azkaban.
+ */
+class SubFlowJob extends StartJob {
+
```
convexquad: Kill extra whitespace line

hadoop-plugin-test/src/main/gradle/positive/groupingWorkflows.gradle
```diff
@@ -0,0 +1,60 @@
...
+      flowDepends 'job1'
+      targets 'sub-job1'
+    }
+
+    noOpJob('job2') {
+    }
+  }
+
+  workflow('nonGroupingFlow') {
+
```
convexquad: Kill extra whitespace line

convexquad:
I am just going to look at the job at the end that is generated with "_end" at the name before I complete the review. Maybe we want to do something a little cleaner there - I'm not sure yet. Thanks for your patience.

convexquad:
Also - can you add a version bump in the gradle.properties file, and add yourself to both the CONTRIBUTORS.md file and also update the VERSIONS.md file accordingly. Please add these to the Pull Request.

convexquad:
Here is the change I want to make:
--Instead of renaming the last noop job in the child flow with "subflowName__end", I want to leave it as "subflowName"
--Then instead of giving the name of the file with type=flow the name "subflowName", give it the name "subflowName__flow" with two underscores (I have established the two underscores as a convention for auto-generated noop jobs used as glue)

The effect of this will be that in the Azkaban UI when the embedded job icon appears that it will have the name "subflowName__flow", which is just a little bit annoying. Then on the main project page you will additionally see the flow "subflowName", which is also not overly annoying.

With the way you had it, on the project page you would see your flow, along with another flow with the name "subflowName__end". I think that is very annoying and also quite confusing. With my proposed changes, it is true that the name of the subflow on the embedded job icon isn't perfect, but that the overall effect is quite positive and nothing will be aggravating for users.

convexquad:
@rride If you don't have time to make the changes I have requested, let me know and I can finish them off! I am eager to get in this great contribution.

rride:
Let me take a look on Monday.

rride:
I made all the requested minor changes and removed __end from a name of the noop job.

However, I can't figure out how to add __flow to sublow's name. Something like subFlowJob = factory.makeSubFlowJob(name + "__flow") is not working. Also, if I rename that, I will have to change target or dependency names of anything depending on that

rride:
Please let me know if you can accept the pull request as is.

convexquad:
@rride Thanks for making the changes! I think what I am going to do is to commit your current PR, and then I'll make the minor fixes to get in the "_flow" at the end of the subflow name. The current approach you mentioned (using factory.makeSubFlowJob(name + "__flow")) is correct - I am sure it is just something minor that is preventing the name from coming out right.

convexquad:
+1 Approved

PR #131, opened 2017-01-02
LIHADOOP-25566: Update Dr.Elephant URL displayed by azkabanFlowStatus task

pranayhasan:
Updated Dr. Elephant link to internal go link


pranayhasan:
There shall be an update to the Dr.E link soon. Also, we realized that its better to avoid using the go link so that it works for other instances as well Eg: holdemaz0[1-6]. Keeping this PR on hold until we update the link.

PR #132, opened 2017-01-06
Check for trailing slash in AzkabanUrl when running azkabanUpload task

pranayhasan:


convexquad:
+1. I will commit this once I regain write access to the project.

PR #135, opened 2017-02-09
LIHADOOP-26251 Include check for classpath property in javadoc when setting cluster provided configuration

pranayhasan:


hadoop-plugin/src/main/groovy/com/linkedin/gradle/hadoop/HadoopPlugin.groovy
```diff
@@ -94,7 +94,9 @@ class HadoopPlugin implements Plugin<Project> {
 
       // Add clusterProvided jars on the classpath, this will not add transitive dependencies of each jar
       project.sourceSets.main.compileClasspath += clusterProvided
-      project.javadoc.classpath += clusterProvided
+      if (project.javadoc.hasProperty("classpath")) {
```
ethankhall: Do you want `project.tasks.javadoc`?

pranayhasan: @ethankhall javadoc task according to gradle is used for generating documentation for classes. I don't know if that's being done here by setting a classpath's config here. Let me know if having this makes sense. Removing this doesn't seem to have making any difference.

convexquad: This code is inside a project.getPlugins().withType(JavaPlugin) { ... } block, so it is safe to assume the javadoc task is present on the project. @pranayhasan instead of project.javadoc (which will indeed find the task), let's indeed follow the convention of using either project.tasks.javadoc or project.tasks['javadoc'], which makes it very explicit that you are looking up the task and not some other kind of extension property.

pranayhasan:
@nntnag17 @ethankhall @convexquad Thanks for the review!

PR #136, opened 2017-02-15
Expand on the list of file types automatically excluded from the sour…

convexquad:
…ces zip to exclude archive types


convexquad:
I probably should have made this a long time ago. Tested successfully with a SNAPSHOT build.

PR #138, opened 2017-03-13
LIHADOOP-25519

nntnag17:
ligradle tasks to test the flow on Azkaban


nntnag17:
@pranayhasan What you said makes perfect sense. Currently, we'll enable this feature for very few of our customers so there won't be any explosion of test projects in Azkaban. We'll fix this issue before the GA. I'll also talk to the Azkaban team to understand if the purger can be a part of their new design.

PR #140, opened 2017-03-31
Incrementing version to 0.11.15 for LinkedIn internal builds

nntnag17:


nntnag17:
Minor change. Self approving.

PR #142, opened 2017-04-06
Remove unfinished Oozie Hadoop DSL elements to simplify the long-term…

convexquad:
… maintenance overhead


convexquad:
@nntnag17 and @pranayhasan can you take a look at this PR? I think we should remove the unfinished Apache Oozie Hadoop DSL compiler that I started 1.5 years ago, since we will probably never finish it. This will simplify the long-term maintenance of the project.

Note that I think we should keep the Oozie command tasks written by @nntnag17 during that time, since they could still be useful to someone using Oozie from the open-source community.

convexquad:
I need to publish an emergency Pull Request to rename the Azkaban test extension methods and tasks from Anant's recent enhancements. I don't want to lose this PR, so I am just going to commit it. If anybody has some feedback on it, we can always fully or partially revert the commit.

PR #143, opened 2017-04-08
Rename conflicting Azkaban test extension and task names to not collide

convexquad:
This is an emergency PR to rename tasks and extension methods that collide with names elsewhere. For example, the "test" extension method collides with the "test" configuration provided by the Gradle Java plugin.

@nntnag17 Although I am going to commit this immediately to unblock our PCX build at LinkedIn, can you take a look at this and re-rename anything I already renamed to names that you like? I just chose the first name that came to mind. Additionally, please check my changes and see if I broke other parts of your code.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,5 +10,3 @@
 bin
 build
 ligradle
-oozie-jaxb/schema
-oozie-jaxb/src/main/java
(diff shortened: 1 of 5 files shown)


convexquad:
Committing to unblock our PCX builds

nntnag17:
Thanks @convexquad  for doing this! I should've used a different name for the test extension.

PR #144, opened 2017-04-11
Add supports for a `required parameters` check

jianlingzhong:
Jobs can now use a "required parameters: []" section to specify any parameters that are required to be set in order for the job to run correctly in production. A static checker is also added so that if any of the required parameters is not set, the DSL static checker will fail.

### DIFF
--- a/hadoop-plugin/src/main/groovy/com/linkedin/gradle/hadoopdsl/checker/RequiredParametersChecker.groovy
+++ b/hadoop-plugin/src/main/groovy/com/linkedin/gradle/hadoopdsl/checker/RequiredParametersChecker.groovy
@@ -1,0 +1,64 @@
+/*
+ * Copyright 2015 LinkedIn Corp.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.linkedin.gradle.hadoopdsl.checker
+
+import com.linkedin.gradle.hadoopdsl.job.Job
+import com.linkedin.gradle.hadoopdsl.BaseStaticChecker
+import com.linkedin.gradle.hadoopdsl.Workflow
+
+import org.gradle.api.Project
+
+class RequiredParametersChecker extends BaseStaticChecker {
+  /**
+   * Constructor for the requiredParametersChecker.
+   *
+   * @param project The Gradle project
+   */
+  RequiredParametersChecker(Project project) {
+    super(project);
+  }
+
+  void checkRequiredParameters(Job job, List<String> flowPath) {
+    if (!job.requiredParameters.isEmpty()) {
+      job.requiredParameters.each { para ->
+        if (!job.jobProperties.containsKey(para)) {
+          String jobPath = flowPath.join('/') + '/' + job.name;
+          project.logger.lifecycle("RequiredParametersChecker ERROR: Job ${jobPath} requires parameter ${para}, but it is not set.");
+          foundError = true;
+        }
+      }
+    }
+  }
+
+  @Override
+  void visitWorkflow(Workflow workflow) {
+    List<String> flowPath = [workflow.name];
+    visitWorkflow(workflow, flowPath);
+  }
+
+  void visitWorkflow(Workflow workflow, List<String> flowPath) {
+    workflow.workflows.each { Workflow flow ->
+      flowPath.push(flow.name);
+      visitWorkflow(flow, flowPath);
+    }
+
+    workflow.jobs.each { Job job ->
+      checkRequiredParameters(job, flowPath);
+    }
+
+    flowPath.pop();
+  }
+}
--- a/hadoop-plugin/src/main/groovy/com/linkedin/gradle/hadoopdsl/job/Job.groovy
+++ b/hadoop-plugin/src/main/groovy/com/linkedin/gradle/hadoopdsl/job/Job.groovy
@@ -48,6 +48,7 @@ class Job {
   Map<String, Object> jobProperties;
   Map<String, String> reading;
   Map<String, String> writing;
+  Set<String> requiredParameters;
 
   /**
    * Base constructor for a Job.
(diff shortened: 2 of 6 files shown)


ahoyleo:
This will be a very useful feature to enforce some custom job archetype (without defining a new job type).

Speaking of, can we also extend the "required" check to dsl methods defined in existing job types, maybe through annotations such as the following:

```
JavaProcessJob extends Job {

  @required
  uses(String javaClass) {
    // ...
  }

}
```

jianlingzhong:
@ahoyleo This would very convenient and can potentially replace the current `RequiredFieldsChecker`.

hadoop-plugin/src/main/groovy/com/linkedin/gradle/hadoopdsl/checker/RequiredParametersChecker.groovy
```diff
@@ -0,0 +1,64 @@
...
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.linkedin.gradle.hadoopdsl.checker
+
+import com.linkedin.gradle.hadoopdsl.BaseStaticChecker
+import com.linkedin.gradle.hadoopdsl.Workflow
+import com.linkedin.gradle.hadoopdsl.job.Job
+import org.gradle.api.Project
```
pranayhasan: Can you move _com.linkedin.gradle.hadoopdsl.job.Job_ to the beginning of imports and give a new line before importing _org.gradle.api.Project_ ?

hadoop-plugin/src/main/groovy/com/linkedin/gradle/hadoopdsl/checker/RequiredParametersChecker.groovy
```diff
@@ -0,0 +1,65 @@
...
+        if (!job.jobProperties.containsKey(para)) {
+          String jobPath = flowPath.join('/') + '/' + job.name;
+          project.logger.lifecycle("RequiredParametersChecker ERROR: Job ${jobPath} requires parameter ${para}, but it is not set.");
+          foundError = true;
+        }
+      }
+    }
+  }
+
+  @Override
```
convexquad: Could you add the JavaDoc comment here from the base class for the overridden method?

hadoop-plugin/src/main/groovy/com/linkedin/gradle/hadoopdsl/checker/RequiredParametersChecker.groovy
```diff
@@ -0,0 +1,65 @@
...
+  void visitWorkflow(Workflow workflow) {
+    List<String> flowPath = [workflow.name];
+    visitWorkflow(workflow, flowPath);
+  }
+
+  void visitWorkflow(Workflow workflow, List<String> flowPath) {
+
```
convexquad: Kill the empty line here

hadoop-plugin/src/main/groovy/com/linkedin/gradle/hadoopdsl/checker/RequiredParametersChecker.groovy
```diff
@@ -0,0 +1,65 @@
...
+   *
+   * @param project The Gradle project
+   */
+  RequiredParametersChecker(Project project) {
+    super(project);
+  }
+
+  void checkRequiredParameters(Job job, List<String> flowPath) {
+    if (!job.requiredParameters.isEmpty()) {
+      job.requiredParameters.each {para ->
```
convexquad: Add a space between the open-bracket and the "para" parameter

hadoop-plugin/src/main/groovy/com/linkedin/gradle/hadoopdsl/job/Job.groovy
```diff
@@ -48,6 +48,7 @@ class Job {
   Map<String, Object> jobProperties;
   Map<String, String> reading;
   Map<String, String> writing;
+  HashSet<String> requiredParameters;
```
convexquad: Could you change the declaration of requiredParameters to a Set<String>, but in the constructor, create it as a LinkedHashSet<String>? Using a LinkedHashSet instead of a regular HashSet will preserve the ordering of the requiredParameters (when enumerated). This will make it possible for us to have consistent unit test outputs.

hadoop-plugin/src/main/groovy/com/linkedin/gradle/hadoopdsl/job/Job.groovy
```diff
@@ -329,6 +332,13 @@ class Job {
     }
   }
 
+  @HadoopDslMethod
+  void required(Map args) {
+    HashSet<String> parameters = args.parameters;
+    for(String entry : parameters) {
```
convexquad: Add a space between "for" and the open-paraen

hadoop-plugin/src/main/groovy/com/linkedin/gradle/hadoopdsl/job/Job.groovy
```diff
@@ -329,6 +332,13 @@ class Job {
     }
   }
 
+  @HadoopDslMethod
```
convexquad: Could you add a comment here briefly describing the idea behind required parameters?

convexquad:
Jianling - I'll take care of quickly adding a couple of unit tests, so actually if you can just resolved the issues opened by myself and Pranay, we'll commit this PR.

jianlingzhong:
@pranayhasan, @convexquad, I should have changed all places asked except those we talked about offline. Thanks for taking care of the unit test @convexquad!

hadoop-plugin/src/main/groovy/com/linkedin/gradle/hadoopdsl/job/Job.groovy
```diff
@@ -330,6 +333,23 @@ class Job {
...
+   * Specifying those parameters here provides a clear and explicit expectation for the users of the job.
+   * For example, specifying "required parameters" in a job template,
+   * and set those parameters in a job cloned from this template.
+   * This helps avoid production errors in an early stage, such as those caused by default parameters.
+   *
+   * @param args Args whose key 'parameters' has a list of value specifying the required parameters
+   */
+  @HadoopDslMethod
+  void required(Map args) {
+    LinkedHashSet<String> parameters = args.parameters;
```
convexquad: @jianlingzhong this is something that users set - I don't think that args.parameters will come out as a LinkedHashSet. Users should pass it in as a List<String> or something similar, no?

jianlingzhong: Yes, this will be a list. I was thinking about initializing this as a `LinkedHashSet` using the List the user passed so that it will take care of duplication at initialization. But maybe a List here would be more readable. I'll change that.

PR #145, opened 2017-04-13
Removed automatic include of runtime dependencies into libPath in ZipTask

vanaoff:
Hey, 

I believe it's not necessary to include automatically all runtime dependencies as these could be provided from different place or in fat jar. Automatic including could in such situations result into massive uploading even when it's not necessary. On the other hand it's straightforward to add these dependencies manually when needed.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -99,6 +99,7 @@ The following were contributed by Pratik Satish. Thanks, Pratik!
 * `LIHADOOP-21429 Add HadoopValidatorPlugin with PigValidatorPlugin in its stack`
 
 The following were contributed by Ivan Heda. Thanks, Ivan!
+* `Removed automatic include of runtime dependencies into libPath in ZipTask.`
 * `Added possibility to have Azkaban password in configuration JSON.`
 
 The following were contributed by Alexander Ivaniuk. Thanks, Alexander!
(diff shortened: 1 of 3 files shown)


vanaoff:
Hi @nntnag17, 

thanks for your comment. However, I don't see how this approach will resolve the issue. Configuration `hadoopRuntime` itself extends the `runtime` configuration and this is where the problem comes from. All the `runtime` dependencies propagate through `hadoopRuntime` into zip, because it's hard-coded in this way.

nntnag17:
Hi @vanaoff 

I agree that hadoopRuntime extends the runtime so anything you add to hadoopRuntime or runtime will automatically go to the zip. I'm asking you to create a new configuration called 'localRuntime' which will extend runtime.  Any dependency added to localRuntime won't be added to hadoopRuntime or runtime. 

You can then add the dependencies that should go to the zip to either hadoopRuntime or runtime.

You can add the dependencies that should not go to the zip to localRuntime. Anything added to localRuntime doesn't get added to either runtime or hadoopRuntime, hence these dependencies won't be added to the zip.

I'm pretty sure that this approach would work. Please let me know if there are any issues with this approach.

vanaoff:
Hi @nntnag17,

thank you for clarification. However, this won't help me because there still remain several `compile` dependencies which are then duplicated in fat jar, which I want to use, and `libPath`. I would probably be able to achieve what I need in some hacky way. 

Nevermind, I finally just manually excluded everything from `hadoopRuntime`. Still, I think it would be clearer to have in zip just what is specified instead of `we-do-it-for-you` magic which could, from time to time, get littlebit non-transparent. Especially in gradle.

Many thanks you for your time, I'm closing this pull request.

PR #146, opened 2017-05-08
Rename core-site.xml and explicitly add to Hadoop Configuration

pranayhasan:
Unless explicitly turned off, Hadoop by default specifies two resources, loaded in-order from the classpath:

core-default.xml: Read-only defaults for hadoop.
core-site.xml: Site-specific configuration for a given hadoop installation.
Applications may add additional resources, which are loaded subsequent to these resources in the order they are added.

The IDE's by default add the defined core-site.xml to classpath causing problems while running tests. So, renaming core-site.xml and adding explicitly to classpath avoids the issue.


convexquad:
@pranayhasan Thanks for fixing this!!!

pranayhasan:
@convexquad Thanks for the quick merge!

PR #151, opened 2017-05-19
After evaluating a Hadoop DSL  we should restore the previous current…

convexquad:
… definition set


convexquad:
Minor bugfix for the Hadoop DSL

PR #152, opened 2017-05-19
Downgrade from com.jfrog.bintray plugin 1.7.3 to 1.7 since the latest…

convexquad:
… version is not working for me


convexquad:
No code changes - this only affects publishing to Bintray and the Gradle plugins portal for open-source users

PR #155, opened 2017-05-31
Refactor namespaces for Hadoop Validator code to be more consistent a…

convexquad:
…nd switch to jcenter


convexquad:
Cleaning up a few things from Pratik's intern project last year for the Hadoop Validator and Pig Validator code.

This looks like a lot of changes, but really the main thing is that I just moved the com.linkedin.gradle.hadoopValidator and com.linkedin.gradle.hadoopValidator.PigValidator namespaces to com.linkedin.gradle.validator.hadoop and com.linkedin.gradle.validator.pig. That should be much more maintainable going forward in case we add any more validators. For example, a "Foo" validator would obviously go in the namespace com.linkedin.gradle.validator.foo.

GitHub can't really tell all that well that mostly I just moved files from one location to another, so it's a bit hard to read the Pull Request. However, I would appreciate any review comments.

PR #156, opened 2017-06-02
Adding Tableau Job to Hadoop DSL

reallocf:
Followed example set by @jenniferdai in #139

### DIFF
--- a/hadoop-plugin/src/main/groovy/com/linkedin/gradle/hadoopdsl/checker/RequiredFieldsChecker.groovy
+++ b/hadoop-plugin/src/main/groovy/com/linkedin/gradle/hadoopdsl/checker/RequiredFieldsChecker.groovy
@@ -32,6 +32,7 @@ import com.linkedin.gradle.hadoopdsl.job.PigJob;
 import com.linkedin.gradle.hadoopdsl.job.PinotBuildAndPushJob;
 import com.linkedin.gradle.hadoopdsl.job.SparkJob;
 import com.linkedin.gradle.hadoopdsl.job.SqlJob;
+import com.linkedin.gradle.hadoopdsl.job.TableauJob;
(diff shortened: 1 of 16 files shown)


hadoop-plugin/src/main/groovy/com/linkedin/gradle/hadoopdsl/job/TableauJob.groovy
```diff
@@ -0,0 +1,94 @@
...
+   * Builds the job properties that go into the generated job file, except for the dependencies
+   * property, which is built by the other overload of the buildProperties method.
+   * <p>
+   * Subclasses can override this method to add their own properties, and are recommended to
+   * additionally call this base class method to add the jobProperties correctly.
+   *
+   * @param parentScope The parent scope in which to lookup the base properties
+   * @return The job properties map that holds all the properties that will go into the built job file
+   */
+  Map<String, String> buildProperties(NamedScope parentScope) {
```
convexquad: You can remove this override, since it doesn't customize the logic for adding properties

hadoop-plugin/src/main/groovy/com/linkedin/gradle/hadoopdsl/checker/RequiredFieldsChecker.groovy
```diff
@@ -224,6 +225,14 @@ class RequiredFieldsChecker extends BaseStaticChecker {
   }
 
   @Override
+  void visitJob(TableauJob job) {
+    if (job.workbookName == null) {
```
cqgao: Add ` || job.workbookName.isEmpty()` to the condition?

reallocf: That seems like a good idea. I know I handle it in my job type code, but handling it in the DSL would be more immediate feedback for users. Thanks!

PR #161, opened 2017-06-09
Minor fix to the test plugin

nntnag17:
If the task buildFlowsForTesting on the subproject and buildSourceZip on the root project run at the same time, buildSourceZip is not able to find some files since they are deleted by buildFlowsForTesting. This PR fixes the issue.


nntnag17:
Hey @convexquad, I've addressed your comments. Please approve

convexquad:
Approved

PR #162, opened 2017-06-09
Hadoop DSL HdfsWaitJob without implementation

ncowanli:
Adding support for the HdfsWait job type.


convexquad:
We discussed over LinkedIn email the separate issue and agreed to go with things as they are now

PR #175, opened 2017-06-19
Add optional skipProfile parameter to the Hadoop DSL applyUserProfile…

convexquad:
… method


convexquad:
Requested by Alexandre Patry - tested successfully with a SNAPSHOT build

Issue #115, opened 2016-10-13
Simpler Hadoop DSL multi-grid builds


convexquad:
I have observed that users struggle in using Hadoop DSL `hadoopClosure` and `namespace` language features to create multi-grid Hadoop DSL workflows.

I have an idea to improve this situation. Essentially, I will provide a new Hadoop DSL mechanism in which you specify multiple definition sets against which to evaluate the Hadoop DSL. The `buildAzkabanFlows` task will then evaluate the Hadoop DSL for the first definition set, then clear its state, re-evaluate the Hadoop DSL for the second definition set, then clear its state, etc. After each re-evaluation of the Hadoop DSL, the compiled output will be written to a unique output location.

Basically, we'll evaluate the Hadoop DSL for each `definitionSet`, clearing it in between, and writing the compiled output to a different location each time.

The advantage of this is that it will be incredibly simpler for users from the perspective of the mental model they need. They will just use `lookupDef` to lookup any values that are different for different grids. Users will not have to use `hadoopClosure` or `namespace` language constructs for multi-grid builds (although you can still use these features).

The disadvantage of this is that it will be slower as it will re-evaluate all of your Hadoop DSL for each `definitionSet`. In addition, you will get WARNING messages from the Hadoop DSL static checker for each re-evaluation. You will also have a lot more compiled output files (a very minor concern).

```
// It will look something like this. In your build.gradle you will have:
apply from: 'src/main/gradle/definitionSets.gradle'  // First, declare your definition sets

// Now customize your Hadoop DSL build
hadoopDslBuild {
  buildPath "azkaban"

  apply files: [
    'src/main/gradle/workflows1.gradle',
    'src/main/gradle/workflows2.gradle',
    'src/main/gradle/common.gradle'
  ]

  definitionSets: ['holdem', 'war']
}

// Now declare that you want to build the Hadoop DSL when you run your build
build.dependsOn buildAzkabanFlows

// When you build, you will have the following output:
//   ${projectDir}/azkaban/holdem
//   ${projectDir}/azkaban/war

// Now easily declare your Hadoop zips for each grid
hadoopZip {
  zip("azkabanHoldem") {
    from "${projectDir}/azkaban/holdem"
  }
  zip("azkabanWar") {
    from "${projectDir}/azkaban/war"
  }
}
```

convexquad:
[...]

PR #182, opened 2017-06-23
First commit for Hadoop automatic mode builds. This first commit auto…

convexquad:
…matically builds the Hadoop DSL for every definition set in the given path.


convexquad:
This is my first cut at automatic Hadoop Plugin builds. With this PR, in your build.gradle you just say:
````
hadoopZip {
  automatic { }
}
```
and it automatically sets the hadoopZip libPath="lib" and sets a build task dependency on building the Hadoop Zips and compiling the Hadoop DSL. For every Gradle file in the folder src/main/definitions, it will rebuild the Hadoop DSL with that definition set and the user's Hadoop DSL profile. Tested successfully with the (internal to LinkedIn) metronome-examples multiproduct.

Next - automatically add one Hadoop zip per definitionSet file and add the compiled Hadoop DSL files for each definition set to the proper zip.

convexquad:
First commit for #115

PR #186, opened 2017-06-29
Adding exactPath clone property for hdfsWaitJob

ncowanli:


convexquad:
Good catch! Sorry I didn't catch it in my review

PR #192, opened 2017-07-27
Adding additional condition for TableauJob creation

reallocf:
Checking to verify workbookName isn't empty on top of being present.

Thanks @cqgao for the suggestion!


pranayhasan:
Thanks Charlie!

Issue #193, opened 2017-10-03
Convert flows to .yml along with .job for Azkaban Flow 2.0


reallocf:
We plan on adapting the HadoopDSL to output two YAML files - .flow and .project - instead of .job/.properties files. Azkaban will read a YAML (.flow) file for each flow as Flow 2.0 is designed and released. The .project file will be used to define Project-level properties, some of which are currently only configurable through the UI (like project permissions).

This allows for easier to define flow-level properties (such as schedules and data-availability based triggers) as well as easier comprehension when trying to understand the generated .zip (job files themselves are hard to read as a cohesive flow unit, especially in large projects).

Note that the hadoopDSL will always able to be configured to output .job files to allow for backward compatibility with old Azkaban versions.

@jamiesjc is in charge of Flow 2.0 and should feel free to add more info :smiley:

reallocf:
I've built a local working version. It still requires some tweaking, but I should be making related commits soon. I plan on integrating with this repo in a few PRs:

1. The YamlJob and YamlWorkflow objects that are used to minimize the Job and Workflow objects for easy snakeyaml conversion. Properties/Property Sets are collapsed into the YamlWorkflow object under its configs section. Also, the PR will include the YamlCompiler object that actually builds and writes the Yaml file. None of these files will be "hooked in" to the codebase.

2. Tests for Yaml conversion to ensure expected output and avoid future regression. This will include unit testing for each of the above objects as well as a few end-to-end tests.

3. Wiring that allows users to decide if they want to output .job/.properties files or a .flow (YAML) file. It will expose an API in gradle. The default will be .job/.properties.

Future TODOs:

- After having a controlled roll-out at LinkedIn, flip the switch so .flow becomes the default option.
- Evaluate sorting the generated nodes for flows/subflows into a linearization of the ultimate Azkaban DAG for easier YAML readability.

pranayhasan:
@reallocf Is having the controlled roll-out plan still a TODO? If not, can you give more details on how we enable it to few users? 
Currently, the Hadoop Plugin has two sub-projects **hadoop-plugin** and an internal **li-hadoop-plugin** specific to Linkedin's use-case. So, you need to hide the custom rollout plan to li-hadoop-plugin. If you're planning to use any internal application specific to Linkedin for Experimentation/AB testing, we can have an offline sych on how we can achieve this.

reallocf:
[...]

PR #196, opened 2017-11-17
First PR for Yaml output for Flow 2.0

reallocf:
This PR introduces the YamlCompiler, YamlWorkflow, YamlJob, and YamlProject objects and a suite of unit tests to cover them.

More details here: #193


hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/yaml/YamlCompiler.groovy
```diff
@@ -0,0 +1,50 @@
...
+
+  /**
+   * Write the YamlWorkflow and YamlProject yaml files into the proper directory
+   *
+   * @param parentDirectory The directory where the Yaml Workflow will be written to
+   * @param workflowName The name of the workflow being output
+   */
+  void dumpYaml(String parentDirectory, String workflowName, String projectName) {
+    def initYamlFile = { fileName ->
+      return new FileWriter(new File(parentDirectory, fileName));
```
jamiesjc: Do we need to close the writer after done?

hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/yaml/YamlProject.groovy
```diff
@@ -0,0 +1,24 @@
...
+ * Representation of a project in flow.project file for Azkaban Flow 2.0
+ */
+class YamlProject {
+  Double azkabanFlowVersion;
+
+  /**
+   * Constructor for YamlProject
+   */
+  YamlProject() {
+    azkabanFlowVersion = 2.0;
```
jamiesjc: Shall we define the constant "2.0" as a static final field?

hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/yaml/YamlWorkflow.groovy
```diff
@@ -0,0 +1,178 @@
...
+   *
+   * @param workflow The workflow to be converted into Yaml
+   * @param parentScope The parent scope of the workflow
+   * @param subflow Boolean regarding whether or not the workflow is a subflow
+   */
+  YamlWorkflow(Workflow workflow, NamedScope parentScope, boolean subflow) {
+    // Include name/type in yaml only if the workflow is a subflow
+    name = subflow ? workflow.name : null;
+    type = subflow ? "flow" : null;
+    dependsOn = workflow.parentDependencies.toList();
```
jamiesjc: What's the difference between parentDependencies and launchDependencies? Why use parentDependencies as the dependsOn list here?

jamiesjc: Today, DSL generates "dependencies", which means the children nodes. But in the new YAML file, the generated "dependsOn" list defines the parents. Is this change reflected somewhere in the code?

jamiesjc:
Is it possible to also add test cases to verify yaml file generated from workflow gradle? Something similar to:
hadoopdsl/linkedin-gradle-plugin-for-apache-hadoop/hadoop-plugin-test/src/main/gradle/positive/workflows.gradle

hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/yaml/YamlWorkflow.groovy
jamiesjc: Never mind, just figured out the difference. "flowDepends" defines parentDependencies, "target" defines launchDependencies.

hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/yaml/YamlWorkflow.groovy
```diff
@@ -0,0 +1,178 @@
...
+      result.add((new YamlWorkflow(subflow, subflow.scope, true)).yamlize());
+    }
+    // Remove all LaunchJobs and StartJobs from dependencies of other jobs
+    workflow.jobsToBuild.each { Job job ->
+      if (job.class == LaunchJob.class || job.class == StartJob.class) {
+        result.each { node ->
+          if (node["dependsOn"]) {
+            node["dependsOn"].remove(job.name);
+            // Remove dependsOn key/val pair from node if dependsOn is now empty
+            if (!node["dependsOn"]) {
```
jamiesjc: node["dependsOn"].isEmpty?

hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/yaml/YamlWorkflow.groovy
```diff
@@ -0,0 +1,178 @@
...
+
+  /**
+   * Create the workflow config from the properties in the namespace
+   *
+   * @param workflow The workflow being constructed
+   * @param parentScope The parent scope of the workflow being constructed
+   * @param subflow Boolean regarding whether or not the workflow is a subflow
+   * @return result The map of all properties associated with the workflow
+   */
+  private static Map buildConfig(Workflow workflow, NamedScope parentScope, boolean subflow) {
```
jamiesjc: How do you handle the scenario where the same property is defined in both the current workflow and the global properties? Should the current flow properties override the global properties?

nntnag17:
Hi Charlie,

I and Pranay will have a look at the code once we have the design review done internally. I think we should also hold off any further changes till the design is reviewed and finalized.

reallocf:
Hey Anant,

Thank you! I think the Flow 2.0 design is mostly complete. @jamiesjc has the latest updates. We'll get back to you with more info next week after thanksgiving. I think Jamie is off through late next week, so might not be until the week after.

Or do you think the hadoopDSL upgrade deserves its own design doc? We haven't thought it necessary up to this point because many of the details are described in Jamie's doc.

hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/yaml/YamlCompiler.groovy
reallocf: Yes - I think we should close the writer. I'll tweak the PR.

In terms of handling the IOException - Groovy doesn't force us to. If the exception is thrown then the execution would fail and a related IOException would be thrown.

It would be nice to attach additional information to this IOException - but no logger is present at this time. Should I just write to stdout? I imagine that would be appropriate for the hadoopDSL.

hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/yaml/YamlProject.groovy
reallocf: "static" isn't an allowed modifier for variables in Groovy. But I can make it final - I'll do so. Thanks!

hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/yaml/YamlWorkflow.groovy
reallocf: That sounds right to me - when I was messing with is parentDependencies seemed like the right one to use and launchDependencies didn't seem necessary, so I went with parentDependencies.

reallocf:
@jamiesjc I'll be adding a number of end-to-end tests validating proper yaml conversion in the next PR. I'm doing it then because the PR will include the infrastructure necessary to manually specify which output (.flow/.project or .job/.property) the user will want to use.

This infrastructure must be in place in order to properly test the entire pipeline (the tests will specify that they want the yaml outputs)

hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/yaml/YamlWorkflow.groovy
reallocf: Yes - I think this is clearer. Thank you!

hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/yaml/YamlWorkflow.groovy
reallocf: I think in the current implementation the global property overrides the workflow property. This doesn't sound right to me, so I'll change. I'll also include a test to confirm this suspicion. Thanks for thinking of it?

Do you agree that workflow property should override global property? We can do either.

hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/yaml/YamlProject.groovy
```diff
@@ -4,13 +4,12 @@ package com.linkedin.gradle.azkaban.yaml;
  * Representation of a project in flow.project file for Azkaban Flow 2.0
  */
 class YamlProject {
-  Double azkabanFlowVersion;
+  final Double azkabanFlowVersion = 2.0;
```
jamiesjc: What about adding the constant 2.0 in AzkabanConstants class?
Something like `static final double AZK_FLOW_VERSION = 2.0`? And assign AZK_FLOW_VERSION to the variable azkabanFlowVersion here?

hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/yaml/YamlWorkflow.groovy
jamiesjc: Yes, I agree with you that workflow property should override global property.

hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/yaml/YamlWorkflow.groovy
```diff
@@ -0,0 +1,180 @@
...
+  List nodes;
+
+  /**
+   * Construct YamlJob from Job and the Job's parent scope
+   *
+   * @param workflow The workflow to be converted into Yaml
+   * @param parentScope The parent scope of the workflow
+   * @param subflow Boolean regarding whether or not the workflow is a subflow
+   */
+  YamlWorkflow(Workflow workflow, NamedScope parentScope, boolean subflow) {
```
jamiesjc: Do you think `isSubflow` better represents a boolean value than `subflow`?

hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/yaml/YamlProject.groovy
reallocf: Good idea - I didn't notice the AzkabanConstants class. Thanks!

hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/yaml/YamlWorkflow.groovy
reallocf: Perhaps, but in the AzkabanDslCompiler class a boolean named `subflow` is used (in the visitWorkflow method). So there's a precedent for using `subflow`.

That being said, `isSubflow` would make the ternary statements more readable. Yeah, I think it's worth changing to `isSubflow`. Thank you!

hadoop-plugin/src/test/groovy/com/linkedin/gradle/azkaban/yaml/YamlProjectTest.groovy
```diff
@@ -0,0 +1,14 @@
...
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+
+class YamlProjectTest {
+  @Test
+  public void TestYamlProject() {
+    Map yamlizedProject = new YamlProject().yamlize();
+
+    assertEquals(2.0, (float) yamlizedProject["Azkaban-Flow-Version"], 0);
```
jamiesjc: `assertEquals(AZK_FLOW_VERSION, yamlizedProject["Azkaban-Flow-Version"]);` ?

hadoop-plugin/src/test/groovy/com/linkedin/gradle/azkaban/yaml/YamlWorkflowTest.groovy
```diff
@@ -0,0 +1,167 @@
...
+    when(mockSubflow.parentDependencies).thenReturn(["testFlow"].toSet());
+    when(mockSubflow.properties).thenReturn([]);
+    when(mockSubflow.jobsToBuild).thenReturn([]);
+    when(mockSubflow.flowsToBuild).thenReturn([]);
+  }
+
+  @Test
+  public void TestSimpleYamlWorkflow() {
+    Map yamlizedWorkflow = new YamlWorkflow(mockWorkflow, mockNamedScope, false).yamlize();
+    assertEquals(false, yamlizedWorkflow.containsKey("name"));
```
jamiesjc: Consider using assertFalse() directly?

hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/yaml/YamlWorkflow.groovy
```diff
@@ -0,0 +1,180 @@
...
+      if (job.class != LaunchJob.class && job.class != StartJob.class) {
+        result.add((new YamlJob(job, parentScope)).yamlize());
+      }
+    }
+    // Add all subflows
+    workflow.flowsToBuild.each { Workflow subflow ->
+      result.add((new YamlWorkflow(subflow, subflow.scope, true)).yamlize());
+    }
+    // Remove all LaunchJobs and StartJobs from dependencies of other jobs
+    workflow.jobsToBuild.each { Job job ->
```
jamiesjc: I think we can keep the lauchJobs and startJobs for now before the new DAG engine feature is ready. I'm ok with checking this part in, but just want to keep a reminder here that the flow execution might not work.

jamiesjc:
The changes look good to me. Left some minor comments.

hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/yaml/YamlWorkflow.groovy
reallocf: Makes sense to me! Thanks for pointing it out. We can refactor this in the future should we want to deploy before the DAG engine is completed.

reallocf:
I've already prepared a second commit that I'll send in another PR for integrating this change with the hadoopDSL, adding the ability for users to specify if they want to generate YAML or not, and adding integration tests.

The only meaningful thing that is different in that PR is: In YamlCompiler, visitProject() is guaranteed to only be called once per execution of the hadoopDSL, instead of once for every visited ScopeContainer. The infrastructure to make that happen is introduced in the new PR, so I didn't include it here.

reallocf:
@nntnag17 if you or somebody from your team could take a look at this over the next few days, I'd really appreciate it! :smiley:

hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/yaml/YamlCompiler.groovy
```diff
@@ -0,0 +1,139 @@
+package com.linkedin.gradle.azkaban.yaml
```
nntnag17: missing semicolon.

hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/yaml/YamlProject.groovy
```diff
@@ -0,0 +1,26 @@
...
+  /**
+   * Constructor for YamlProject
+   */
+  YamlProject(String name) {
+    this.name = name;
+  }
+
+  /**
+   * @return Map detailing exactly what should be printed in Yaml
+   */
```
nntnag17: Missing @Override here and at other classes which implement YamlObject

nntnag17:
Hey @reallocf ,

I've started the review. Will finish it soon.

hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/yaml/YamlProject.groovy
reallocf: Nice catches - thank you!

pranayhasan:
I too have started it. Would complete my review over the weekend.

reallocf:
Anything else you need from me on this PR? Thanks for everybody's help!

PR #197, opened 2017-11-20
Bump to Gradle 4.1

pranayhasan:
Starts a new series 0.14.x with Gradle version increment from 2.13 to 4.1
Users are advised to avoid usage of left shift operator and instead use doLast as an alternative.

**TODO**
- [x] Gradle wrapper updated from maven repository
- [x] Test Eclipse and Idea auto-completion, if functional
- [x] Test azkaban plugin tasks


pranayhasan:
Hey @nntnag17, I've completed verifying the TODO list. Can you please review?

PR #198, opened 2017-12-12
Addressed a bug in ready status in Azkaban client. Cleaned up the codenarc errors in azkaban-client.

pranayhasan:


pranayhasan:
It's to make style check happy and make azkaban-client as an example subproject to maintain coding style for the rest of the project. Let me know if that doesn't makes sense in here.

convexquad:
Just FYI with some historical context - when we started the Hadoop Plugin, the LinkedIn Gradle Groovy plugins did not automatically run the Groovy style check task as a part of `gradle build`, and thus I wrote the initial code in a varied style.

Unfortunately one of the main language drawbacks (or benefit as you see it) is that it is possible / inevitable to write Groovy code in a varied style. Thus if you run the Groovy style check task on the code now, you will get hundreds of warnings.

That's why in the current Gradle setup, the style check task is explicitly disabled... it would be a bit of an investment to re-enable the style check task and fix all the warnings (although some of the more picky style rules could be individually disabled).

Hope you guys are well!

pranayhasan:
Thanks Alex for the context. 
I hope you're doing well. :)

Issue #110, opened 2016-09-28
Ability to upload to multiple Azkaban instances


convexquad:
Update the .azkabanPlugin.json file so that it supports a list of named Azkaban instances. One of them will be the "default" instance. Currently, the file looks like:

```
{
    "azkabanUrl": "https://ltx1-holdemaz01.grid.linkedin.com:8443",
    "azkabanProjName": "abain-xgboost-demo",
    "azkabanZipTask": "azkabanHadoopZip",
    "azkabanValidatorAutoFix": "true",
    "azkabanUserName": "abain"
}
```

It should look something like:

```
{
  {
    // Instance that has no name is the default one
    "azkabanUrl": "https://myDefaultAzkabanInstance.linkedin.com:8443",
    "azkabanProjName": "myAzkabanProject",
    "azkabanZipTask": "azkabanHadoopZip",
    "azkabanValidatorAutoFix": "true",
    "azkabanUserName": "abain"
  },
  {
    "azkabanName": "secondInstance"   
    "azkabanUrl": "https://secondInstance.linkedin.com:8443",
    "azkabanProjName": "myOtherProject",
    "azkabanZipTask": "azkabanHadoopZip",
    "azkabanValidatorAutoFix": "true",
    "azkabanUserName": "abain"
  },
  ...
}
```

This enhancement needs to be done in such a way that it is backwards compatible with the existing format of the .azkabanPlugin.json file, or perhaps it notices when the file is in the old format and automatically updates it.

In the azkabanUpload task you would pass -PazkabanName=secondInstance on the command line to use the second config. In the interactive mode if the user makes any edits to the values we would ask them if they want to save the changes to the same instance, or to created a new named instance and save the changes to that instance.

pranayhasan:
@convexquad How often does users have this user-case? 
.azkabanplugin.json is project dependent, so we have one file each for a project. I believe most of our users upload to test cluster or prod. Once users are done testing, they change it to prod and they are done.
It's a good enhancement, but are we overdoing it? I think it would confuse a lot of new users to adopt to this change. Feel free to put your argument.

convexquad:
@pranayhasan Two people have asked me about it so far. In general, LinkedIn is heading to a future where it has many different clusters, so I do think that this is a useful enhancement. However, I agree that we should be careful about overdoing it and be careful about providing features that are not actually that helpful.

[...]

Issue #90, opened 2016-08-20
Enable azkabanUpload task to validate session before uploading


pranayhasan:
For an invalid session, the azkabanUpload task performs uploading the zip twice where the first POST request fails and the further request succeeds. Thus it takes twice the time necessary for uploading the zip, if the session is invalid. A quick fix is to check each time if the session is valid and then proceed with upload. @convexquad @nntnag17 @akshayrai Please let me know if there is a better way, if any.

convexquad:
@pranayhasan How would you check if the session is valid? Is there something you can reliably call in Azkaban to check?

pranayhasan:
It can be made by calling azkaban API like [fetchprojectflows](http://azkaban.github.io/azkaban/docs/latest/#api-fetch-flows-of-a-project) which just needs project name as a parameter.

convexquad:
I think that API endpoint sounds good. In addition to verifying if the user has logged in, it might be nice to use this to verify that the project they specified exists. +1 to this enhancement

convexquad:
@pranayhasan Ok this happened to me today when I was uploading, and I needed to enter my password. I agree that this would be a good enhancement.

pranayhasan:
#98

PR #199, opened 2018-01-06
session ids are stored separately for each azkaban server

mtrna:
Hi! I would like to improve the handling of the session ids a bit.
current master: only one session id for the last Azkaban contacted is stored
proposed change: each Azkaban has its separate file, where session id and timestamp of its retrieval is stored
expected benefits: Azkaban sessions are preserved when working with multiple Azkabans; old session ids are not used; users should see less of unnecessary authentication errors.

### DIFF
--- a/hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/AzkabanHelper.groovy
+++ b/hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/AzkabanHelper.groovy
@@ -116,7 +116,7 @@ class AzkabanHelper {
       }
 
       String sessionId = jsonObj.getString("session.id");
-      saveSession(sessionId);
+      saveSession(azkabanUrl, sessionId);
       return sessionId;
     }
     finally {
@@ -412,19 +412,27 @@ class AzkabanHelper {
   /**
    * Reads the Azkaban session ID from the session.properties file in the ~/.azkaban directory.
    *
+   * @param azkabanUrl Azkaban URL
    * @return The Azkaban session ID
    */
-  static String readSession() {
-    File file = new File(System.getProperty("user.home") + "/.azkaban/session.properties");
-    String sessionId = null;
+  static String readSession(String azkabanUrl) {
+    String path = System.getProperty("user.home") + "/.azkaban/" + filesystemFriendly(azkabanUrl) + ".properties"
+    File file = new File(path);
+    logger.lifecycle("Searching for properties on path " + path + "\n");
     if (file.exists()) {
+      String sessionId = null;
       file.withInputStream { inputStream ->
+        logger.lifecycle("Properties found on path " + path + "\n");
         Properties properties = new Properties();
         properties.load(inputStream);
         sessionId = properties.getProperty("sessionId");
       }
+      logger.lifecycle("Session found. Session ID " + sessionId + " deemed valid.");
+      return sessionId;
+    } else {
+      logger.lifecycle("No properties found on path " + path);
+      return null
     }
-    return sessionId;
   }
 
   /**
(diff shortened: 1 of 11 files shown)


hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/AzkabanHelper.groovy
```diff
@@ -412,19 +412,36 @@ class AzkabanHelper {
...
-  static String readSession() {
-    File file = new File(System.getProperty("user.home") + "/.azkaban/session.properties");
-    String sessionId = null;
+  static String readSession(String azkabanUrl) {
+    String path = System.getProperty("user.home") + "/.azkaban/" + filesystemFriendly(azkabanUrl) + ".properties"
+    File file = new File(path);
+    logger.lifecycle("Searching for properties on path " + path + "\n");
     if (file.exists()) {
+      String sessionId = null;
+      long sessionTimestamp = 0
```
pranayhasan: The assumption of Azkaban's session TTL is unnecessary as the validation is being done by individual instances calling the readSession() method.  If the session's expired the user is asked to enter the password which is already taken care of.

Can you please make changes to remove the use of timestamp and validation for it in here?

Overall, this PR compliments #110 for users using multiple azkaban instances concurrently. I'm looking forward for your changes :)

hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/AzkabanHelper.groovy
```diff
@@ -483,6 +506,7 @@ class AzkabanHelper {
     try {
       file.withWriter { writer ->
         Properties properties = new Properties();
+        properties.setProperty("timestamp", System.currentTimeMillis().toString())
```
pranayhasan: Wouldn't be necessary to set based on my above comments.

hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/AzkabanHelper.groovy
mtrna: Hi @pranayhasan! The sole purpose of the `sessionTimestamp` is to put upper limit on the age of the `sessionId`. If a session is retrieved by the plugin on say Friday and then the plugin is used again on Monday, the `sessionId` is already considered expired and not used anymore.
Use case: this comes very handy on slow networks during the execution of the `azkabanUpload` task: consider that the Azkaban zip is high tens of megabytes in size and we want to upload it after more then a day of no communication with Azkaban. If we try to use sessionId indiscriminately, then we could try to upload a huge file, tens of minutes later this upload would predictably fail (since the sessionId from long ago is not valid anymore), then we would retrieve a fresh sessionId, and then the upload would finally succeed after considerable amount of time. This is not practical.

pranayhasan: Your mentioned use-case is already addressed in #90
Although your mentioned way invalidates an unnecessary call in few cases, I think it's better to keep it simple :)
Let me know if you're not convinced as we can take @nntnag17 's opinion.

mtrna: Hi @pranayhasan ! I am not strongly opposed to removing it, it just makes more sense to me that a sessionId is accompanied by a timestamp. Session to me means a token+lifespan. From the practical impact: it makes difference on a slow line, and the overhead of checking the validity of the token is low.
Sure, let's hear @nntnag17 's or anybody else's opinion to get extra insight. :)

nntnag17: Hi! I think the benefit that we get from adding a timestamp is not huge since #90 already makes a lightweight request to check if sessionId is valid. Also, as a client, we should not make assumptions about the session lifespan which is governed by the server. IMHO, it's better not to add the timestamp to the session id and let #90 handle session validation.

pranayhasan:
@mtrna It would be great if you could add yourself to CONTRIBUTORS.md, version bump gradle.properties and have a corresponding description to VERSIONS.md. Also, it would be helpful if you can rebase your commits on top of the _linkedin/master_ when you're modifying these files as there shall be conflicts.

hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/AzkabanHelper.groovy
pranayhasan: Thanks @nntnag17

mtrna:
@pranayhasan : thank you! the commits were squashed, the PR should be ready for merge now. :)

pranayhasan:
Congrats @mtrna  on your first contribution to the Hadoop Gradle Plugin. I hope there are a lot more to come! :+1:

PR #200, opened 2018-01-24
Second PR for Yaml output for Flow 2.0

reallocf:
This is the second and final PR related to this issue: #193 

This PR builds the interface users interact with to choose whether or not they want to output .flow/.project files or .job/.properties files. As stated in my prior PR (#196) the default is .job/.properties for now.

After this PR is merged (and subsequently released), users can specify that they want to output yaml by adding:
* `generateYamlOutput true` or
* `generateYamlOutput()`

in the hadoop closure within their gradle files.

The change to existing code to make this happen is only ~75 lines, the other 500 lines added are from adding 4 end-to-end tests that confirm basic functionality for Flow 2.0. These tests mirror those done on the Azkaban side to confirm proper uploading.

### DIFF
--- a/hadoop-plugin-test/expectedJobs/basicFlow/basicFlow.flow
+++ b/hadoop-plugin-test/expectedJobs/basicFlow/basicFlow.flow
@@ -1,0 +1,25 @@
+config:
+  flow-level-parameter: value
+nodes:
+- name: basicFlow
+  type: noop
+  dependsOn:
+  - shellEnd
+- name: shellEnd
+  type: noop
+  dependsOn:
+  - shellPwd
+  - shellEcho
+  - shellBash
+- name: shellPwd
+  type: command
+  config:
+    command: pwd
+- name: shellEcho
+  type: command
+  config:
+    command: echo "This is an echoed text."
+- name: shellBash
+  type: command
+  config:
+    command: bash ./sample_script.sh
--- a/hadoop-plugin-test/expectedJobs/basicFlow/basicFlow.flow
+++ b/hadoop-plugin-test/expectedJobs/basicFlow/basicFlow.flow
@@ -4,10 +4,6 @@ nodes:
 - name: basicFlow
   type: noop
   dependsOn:
-  - shellEnd
-- name: shellEnd
-  type: noop
-  dependsOn:
   - shellPwd
   - shellEcho
   - shellBash
--- a/hadoop-plugin-test/expectedJobs/embeddedFlow/embeddedFlow.flow
+++ b/hadoop-plugin-test/expectedJobs/embeddedFlow/embeddedFlow.flow
@@ -1,0 +1,62 @@
+config:
+  flow-level-parameter: value
+nodes:
+- name: embeddedFlow
+  type: noop
+  dependsOn:
+  - shellEnd
+- name: shellEnd
+  type: noop
+  dependsOn:
+  - shellPwd
+  - shellEcho
+  - embeddedFlow1
+- name: shellPwd
+  type: command
+  config:
+    command: pwd
+- name: shellEcho
+  type: command
+  config:
+    command: echo "This is an echoed text."
+- name: embeddedFlow1
+  type: flow
+  config:
+    flow-level-parameter: value
+  nodes:
+  - name: embeddedFlow1
+    type: noop
+    dependsOn:
+    - shellEnd
+  - name: shellEnd
+    type: noop
+    dependsOn:
+    - shellEcho
+    - embeddedFlow2
+  - name: shellEcho
+    type: command
+    config:
+      command: echo "This is an echoed text from embeddedFlow1."
+  - name: shellBash
+    type: command
+    config:
+      command: bash ./sample_script.sh
+  - name: embeddedFlow2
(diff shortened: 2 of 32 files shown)


hadoop-plugin-test/expectedJobs/basicFlow/basicFlow.flow
```diff
@@ -0,0 +1,25 @@
+config:
+  flow-level-parameter: value
+nodes:
+- name: basicFlow
```
jamiesjc: Why do you need to create two noop jobs: basicFlow and shellEnd? Will one be enough?

hadoop-plugin-test/expectedJobs/embeddedFlow/embeddedFlow.flow
```diff
@@ -0,0 +1,62 @@
+config:
+  flow-level-parameter: value
+nodes:
+- name: embeddedFlow
+  type: noop
```
jamiesjc: Same comments as above.

hadoop-plugin-test/expectedJobs/basicFlow/basicFlow.flow
reallocf: Discussed offline. You are right! I removed shellEnd jobs.

hadoop-plugin-test/expectedJobs/embeddedFlow/embeddedFlow.flow
reallocf: Same as above :+1:

reallocf:
@nntnag17 have time to review this today? The actual code is small - it's mostly just tests. Thanks so much!

hadoop-plugin-test/src/main/gradle/positive/embeddedFlow.gradle
```diff
@@ -0,0 +1,69 @@
+buildscript {
+  repositories {
+    jcenter()
+  }
+  dependencies {
+    classpath files("${project.pluginTestDir}/hadoop-plugin-${project.version}.jar", "${project.pluginTestDir}/hadoop-plugin-${project.version}-SNAPSHOT.jar")
+    // Must manually specify snakeyaml for testing, not required for users
+    classpath 'org.yaml:snakeyaml:1.18'
```
nntnag17: Have you confirmed that this version of snakeyaml is in LinkedIn artifactory? The build might fail internally if it's not there.

hadoop-plugin/src/main/groovy/com/linkedin/gradle/hadoopdsl/BaseCompiler.groovy
```diff
@@ -73,6 +73,8 @@ abstract class BaseCompiler extends BaseVisitor implements HadoopDslCompiler {
     visitExtension(plugin.extension);
   }
 
+  abstract void doAtCompilerCreation();
```
nntnag17: Can you add a comment block to explain the function?

hadoop-plugin/src/main/groovy/com/linkedin/gradle/hadoopdsl/BaseCompiler.groovy
```diff
@@ -108,6 +110,8 @@ abstract class BaseCompiler extends BaseVisitor implements HadoopDslCompiler {
       cleanBuildDirectory(file);
     }
 
+    doAtCompilerCreation();
```
nntnag17: Can we name this function more appropriately? The current name sounds like this function should be called in the constructor of each compiler.

hadoop-plugin/src/main/groovy/com/linkedin/gradle/hadoopdsl/BaseNamedScopeContainer.groovy
```diff
@@ -645,6 +645,30 @@ abstract class BaseNamedScopeContainer implements NamedScopeContainer {
...
+
+  /**
+   * Creates the task to toggle between .flow/.project files generation and .job/.properties files
+   * generation.
+   *
+   * @param boolean flag Boolean to which generate_yaml_output is set
+   * @return The value of generate_yaml_output flag - always true
+   */
+  @HadoopDslMethod
+  boolean generateYamlOutput(boolean flag) {
```
nntnag17: If the user passes `generateYamlOutput false` in the workflow scope, what should be the behaviour? I assume that it should not produce YAML output for consistency. Can you also add a test for this?

hadoop-plugin-test/src/main/gradle/positive/embeddedFlow.gradle
reallocf: Good thinking! Yes, I see `snakeyaml-1.18.jar` in artifactory :+1:

hadoop-plugin/src/main/groovy/com/linkedin/gradle/hadoopdsl/BaseNamedScopeContainer.groovy
reallocf: Good idea - I'll add one now.

PR #202, opened 2018-03-03
LIHADOOP-32983 Azkaban CLI Tasks should prompt for password+VIP internal to Linkedin

pranayhasan:
Corrects the password prompt to print "Enter Password+VIP" ONLY inside Linkedin.

### DIFF
--- a/VERSIONS.md
+++ b/VERSIONS.md
@@ -17,6 +17,11 @@ the License.
 Note that the LinkedIn build system occasionally requires that we skip a
 version bump, so you will see a few skipped version numbers in the list below.
 
+0.14.5
+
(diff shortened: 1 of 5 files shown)


VERSIONS.md
```diff
@@ -17,6 +17,11 @@ the License.
 Note that the LinkedIn build system occasionally requires that we skip a
 version bump, so you will see a few skipped version numbers in the list below.
 
+0.14.4
```
arpang: Can you change it 0.14.5? I have already used 0.14.4 for adding AutoTunePigLi job type.

pranayhasan: Yes sure. Once, I push your change I will update and push this.
Thanks!

Issue #194, opened 2017-10-03
Azkaban schedules and data-availability based triggers defined in HadoopDSL


reallocf:
The Azkaban team is working on data-availability based triggers and will be launching that feature with HadoopDSL integration.

These dependencies will be written in .flow files that were introduced in Flow 2.0 (#193). This feature will not be available for users outputting .job/.properties files.

As a side-product of data-availability based triggers, this change will also allow HadoopDSL-defined schedules to be created.

In the future, we're aiming for all jobs and their associated schedules/triggers to be defined in .flow files. This will allow versioning of schedules, which hasn't been possible in the past.

Old .job/.properties files will still be able to be generated in the future for backward compatibility with older Azkaban versions, but won't have these new features.

@chengren311 is leading the data-availability based trigger project on the Azkaban team, and he should feel free to add more if he so chooses :smiley:

prokod:
@reallocf, @chengren311 this is very exciting stuff!
Could you please let users/followers know what is the status on that ? Specifically if I now pull Azkaban 3.43.0 and I deploy it, could I already generate flow 2.0 type flows in a manual way ? (probably only future versions of the Hadoop plugin will have this functionality as latest plugin was released in June 2017)
How close are you to release data-availability based triggers ?

Kudos on the fine work

reallocf:
Hey @prokod - I believe the data-availability based triggers are very close to working! I've been leading the work on implementing them here in the DSL. It expect it will take another few months before they're fully ready due to my attention being elsewhere as GDPR approaches.

@chengren311 should be able to give you more details about the data trigger release though! Last I heard it should be available in Azkaban within the next month :)

chengren311:
@prokod @reallocf  the data trigger feature is under active development and will be released soon. Most part are done except some misc things.  There's a wiki page(https://github.com/azkaban/azkaban/wiki/Azkaban-Data-Trigger) (more things to be added )about the feature if you are interested.

reallocf:
Starting work on the DSL for this in earnest now. Will keep posted with updates here.

reallocf:
This is completed and in the process of being rolled out internally :+1: 
Woo!

PR #212, opened 2018-05-08
DSL Trigger Implementation

reallocf:
As Azkaban rolls out data-availability based triggers, we're updating the DSL so it can generate these triggers.

Note: Since triggers are flow-level, it requires the new Flow 2.0 yaml serialization. You can instruct the DSL to generate yaml by specifying `generateYamlOutput true` in your hadoop closure.

This PR provides a full data triggers implementation. It introduces the Trigger, Schedule, and TriggerDependency objects.
- Trigger objects can be created in any NameScope, but will only be picked up to be output if defined in (or added to) a Workflow. An error will be thrown if more than one Trigger is defined in a Workflow.
- Schedule objects are created inside of Trigger objects and specify the cron value corresponding to when the Trigger will be created in Azkaban. An error will be thrown if more than one schedule is created in a Trigger OR if no schedule is created in a trigger.
- The TriggerDependency abstract object is built to be implemented by many classes so the interface is pluggable. As of this PR there is only one implementation corresponding with the one implementation available in Azkaban when Data Triggers are released - the DaliDependency. Unfortunately, this is a closed-source dependency, so it doesn't offer a great deal of value to the open source community. When the HDFS Dependency is introduced in Azkaban, a corresponding TriggerDependency will also be created here.

Other than the introduction of these objects and the logic to yamlize them in the AzkabanDslYamlCompiler, this PR is primarily checker logic and unit and integration tests. The TriggerChecker is implemented to make sure that Triggers/Schedules/TriggerDependencies are properly defined. The unit tests confirm that the AzkabanDslYamlCompiler compiles the new objects properly. And the integration tests test confirm that the DSL is generating Triggers in Yaml in an expected way and that the TriggerChecker picks up improperly defined Triggers.

More info in #194

### DIFF
--- a/hadoop-plugin-test/src/main/gradle/negative/triggerCheck.gradle
+++ b/hadoop-plugin-test/src/main/gradle/negative/triggerCheck.gradle
@@ -34,10 +34,10 @@ hadoop {
 
     trigger('trigger2') { // two defined triggers throws an ERROR
       maxWaitMins 0 // maxWaitMins less than 1 throws an ERROR
-      schedule('schedule1') {
+      schedule {
         // no schedule value throws an ERROR
       }
-      schedule('schedule2') { // two schedules throws an ERROR
+      schedule { // two schedules throws an ERROR
         value 'abcd' // invalid cron expression for value in schedule throws an ERROR
       }
     }
@@ -52,7 +52,7 @@ hadoop {
   workflow('triggerFail2') {
     trigger('trigger3') {
       maxWaitMins 9999999 // really high maxWaitMins throws a WARN
-      schedule('schedule1') {
+      schedule {
         value '*/30 0 1 ? * *'// really small cron interval throws an ERROR
       }
       daliDependency('dali-data-dep') {
--- a/hadoop-plugin-test/src/main/gradle/negative/triggerCheck.gradle
+++ b/hadoop-plugin-test/src/main/gradle/negative/triggerCheck.gradle
@@ -12,12 +12,10 @@ buildscript {
 
 apply plugin: com.linkedin.gradle.hadoop.HadoopPlugin
 
-// Simple flow for testing trigger creation
-
-// TODO reallocf negative cases for Cheng's negative cases in AZ
+// Negative cases for trigger creation
 
 hadoop {
-  buildPath "jobs/triggerFlow"
+  buildPath "jobs/triggerCheck"
   cleanPath false
 
   generateYamlOutput true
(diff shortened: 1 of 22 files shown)


hadoop-plugin/src/main/groovy/com/linkedin/gradle/hadoopdsl/checker/TriggerChecker.groovy
```diff
@@ -0,0 +1,199 @@
...
+      project.logger.lifecycle("TriggerChecker ERROR: DaliDependency ${daliDependency.name} in Trigger ${trigger.name} in Workflow ${workflow.name} must define a unit.");
+      checkDaliDependencyError = true;
+    }
+    // ERROR if dali dependency has a window less than one.
+    if (daliDependency.params["window"] < 1) {
+      project.logger.lifecycle("TriggerChecker ERROR: DaliDependency ${daliDependency.name} in Trigger ${trigger.name} in Workflow ${workflow.name} must have a window greater than 0.");
+      checkDaliDependencyError = true;
+    }
+    // ERROR if dali dataset dependency has a window less than one.
+    if (daliDependency.params["delay"] < 1) {
```
reallocf: Should be < 0

hadoop-plugin-test/src/main/gradle/positive/triggerFlow.gradle
```diff
@@ -0,0 +1,64 @@
...
+hadoop {
+  buildPath "jobs/triggerFlow"
+  cleanPath false
+
+  generateYamlOutput true
+
+  workflow('triggerFlow') {
+    trigger('trigger') {
+      maxWaitMins 5
+      schedule('schedule') {
```
reallocf: Remove name (?)

hadoop-plugin/src/test/groovy/com/linkedin/gradle/azkaban/AzkabanDslYamlCompilerTest.groovy
```diff
@@ -17,11 +17,14 @@ package com.linkedin.gradle.azkaban;
...
 import com.linkedin.gradle.hadoopdsl.NamedScope;
 import com.linkedin.gradle.hadoopdsl.Properties;
+import com.linkedin.gradle.hadoopdsl.Schedule;
+import com.linkedin.gradle.hadoopdsl.Trigger;
 import com.linkedin.gradle.hadoopdsl.Workflow;
 import com.linkedin.gradle.hadoopdsl.job.Job;
 import com.linkedin.gradle.hadoopdsl.job.LaunchJob;
 import com.linkedin.gradle.hadoopdsl.job.StartJob;
-import com.linkedin.gradle.hadoopdsl.job.SubFlowJob;
+import com.linkedin.gradle.hadoopdsl.job.SubFlowJob
```
reallocf: add semicolon

hadoop-plugin-test/src/main/gradle/negative/triggerCheck.gradle
```diff
@@ -0,0 +1,84 @@
...
+
+    commandJob('shellBash') {
+      uses 'bash ./sample_script.sh'
+    }
+
+    targets 'shellBash'
+  }
+
+  workflow('triggerFail2') {
+    trigger('trigger3') {
```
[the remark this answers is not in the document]
reallocf: So Triggers are one of the objects that can be defined in any BaseNamedScope. The benefit of this is that you can define the trigger in one place (like a global location) then reference it by name it will be added to a workflow.

Because of this, if you create a trigger named (for example) 'trigger1' outside of a workflow, then add it to a workflow with the command
```
addTrigger('trigger')
```
inside of that workflow, referencing by name.

So it does need _some_ name. I guess it doesn't necessarily have to be `trigger3`, but that does make sense given the other trigger names :)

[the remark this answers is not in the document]
chengren311: got it, thanks

PR #213, opened 2018-05-15
Remove unneeded dependency in triggerCheck.gradle

reallocf:
compareKnownOutputFiles test is failing in new test environments (experienced via CRT) because a quartz dependency is being downloaded at this step.

I'm hoping that removing this (unneeded) classpath inclusion will fix it.


chengren311:
please fix the build.

reallocf:
After further investigation, it looks like this approach is wrong. I'm going to try to deploy again - hopefully it was something intermittent.

PR #221, opened 2018-08-01
Handling emergent flows for Flow 2.0

reallocf:
In Flow 1.0, we had an interesting property (called emergent flows) where users could write .properties files to a separate directory than the original workflow directory during compile time, then include them both at zip time to create the workflow that is ultimately uploaded. This was nice because you could specify multiple zip files and pick different properties for each zip file in order to have different properties on different clusters. This was a heavily used mechanism for having cluster-specific properties at LinkedIn.

For Flow 2.0 this is less straightforward because the Yaml object is completely generated at compile time. So this PR introduces a similar mechanism to allow properties to be merged into flows at zip time.

The way it works is like this:

If a namespace has NO workflows defined but DOES have properties defined, it writes those properties to a .tempprops file. Properties defined with workflows are expected to be workflow properties.

At zip time, if the zip contains any .tempprops files, it reads them in then reads in all the .flow files and merges them together. It then creates new .flow files with the resulting properties included. The files are named <original-flow-name>_<zip-name>.flow so each zip will create a uniquely named .flow file. The original .flow files and .tempprops files are then not included in the zip while the new .flow files are.

This should create backward compatibility with Flow 1.0 for flows that leverage the emergent flows property.

### DIFF
--- a/hadoop-plugin-test/src/main/gradle/positive/emergentFlow.gradle
+++ b/hadoop-plugin-test/src/main/gradle/positive/emergentFlow.gradle
@@ -50,19 +50,3 @@ hadoop {
     targets 'shellPwd', 'shellEcho', 'shellBash'
   }
 }
-
-hadoopZip {
-  libPath = "lib"
-  base {
-    from { fileTree("jobs/emergentFlow").exclude("otherNamespace").files } {
-      into "."
-    }
-  }
-  zip("emergentFlowTest") {
-    from { fileTree("jobs/emergentFlow/otherNamespace").files } {
-      into "."
-    }
-  }
-}
-
-startHadoopZips.dependsOn buildAzkabanFlows
--- a/hadoop-plugin-test/src/main/gradle/positive/emergentFlow.gradle
+++ b/hadoop-plugin-test/src/main/gradle/positive/emergentFlow.gradle
@@ -1,0 +1,68 @@
+buildscript {
+  repositories {
+    jcenter()
+  }
+  dependencies {
+    classpath files("${project.pluginTestDir}/hadoop-plugin-${project.version}.jar", "${project.pluginTestDir}/hadoop-plugin-${project.version}-SNAPSHOT.jar")
+    // Must manually specify snakeyaml for testing, not required for users
+    classpath 'org.yaml:snakeyaml:1.18'
+  }
+}
+
+apply plugin: com.linkedin.gradle.hadoop.HadoopPlugin
+
+// Flow for testing yaml creation for emergent flows for Flow 2.0
+
+hadoop {
+  buildPath "jobs/emergentFlow"
+  cleanPath false
+
+  generateYamlOutput true
+
+  namespace('otherNamespace') {
+    propertyFile('emergentProperties') {
+      set properties: [
+              'new-parameter' : 'new-value',
+              'flow-level-parameter' : 'value2'
+      ]
+    }
+  }
+
+  workflow('emergentFlow') {
+    propertyFile('properties') {
+      set properties: [
+              'flow-level-parameter' : 'value'
+      ]
+    }
+
+    commandJob('shellBash') {
+      uses 'bash ./sample_script.sh'
+    }
+
+    commandJob('shellPwd') {
+      uses 'pwd'
+    }
+
+    commandJob('shellEcho') {
+      uses 'echo "This is an echoed text."'
+    }
+
+    targets 'shellPwd', 'shellEcho', 'shellBash'
+  }
+}
+
+hadoopZip {
+  libPath = "lib"
+  base {
+    from { fileTree("jobs/emergentFlow").exclude("otherNamespace").files } {
+      into "."
+    }
+  }
+  zip("emergentFlowTest") {
+    from { fileTree("jobs/emergentFlow/otherNamespace").files } {
+      into "."
+    }
+  }
+}
+
+startHadoopZips.dependsOn buildAzkabanFlows
--- a/hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/AzkabanDslYamlCompiler.groovy
+++ b/hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/AzkabanDslYamlCompiler.groovy
@@ -30,10 +30,10 @@ import com.linkedin.gradle.hadoopdsl.job.SubFlowJob;
 import com.linkedin.gradle.hadoopdsl.triggerDependency.DaliDependency;
 import com.linkedin.gradle.hadoopdsl.triggerDependency.TriggerDependency;
 import org.gradle.api.Project;
-import org.yaml.snakeyaml.DumperOptions;
 import org.yaml.snakeyaml.Yaml;
 
 import static com.linkedin.gradle.azkaban.AzkabanConstants.AZK_FLOW_VERSION;
+import static com.linkedin.gradle.util.YamlUtils.setupYamlObject;
 
 /**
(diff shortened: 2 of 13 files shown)


hadoop-plugin/src/main/groovy/com/linkedin/gradle/zip/HadoopZipExtension.groovy
```diff
@@ -204,6 +209,55 @@ class HadoopZipExtension {
...
+        Map<String, String> tempPropsMap = YamlMerge.readInYaml(tempProps);
+        tempPropsList.add(tempPropsMap);
+        tempPropsPathsList.add(tempPropsPath);
+      }
+    }
+    // Merge all .tempprops files into the .flow files and create new .flow files with the results
+    if (!tempPropsList.isEmpty()) {
+      task.source.visit { FileVisitDetails file ->
+        String flowFile = file.path;
+        if (flowFile.matches(/.*\.flow/) && !flowFile.matches(/.*_.*/)) {
```
jamiesjc: Can the name of a flow file before merging also contain underscore? e.g. "jasun_test.flow"

reallocf: Good question - no it cannot. Flow names (and all object names in the DSL in general) can only contain letters, numbers, and dashes.

jamiesjc: I remember the job file names in Flow 1.0 allow using "_". Is it some new requirement for the flow names in Flow 2.0?  Can you show me where the code is? Thanks!

hadoop-plugin-test/src/main/gradle/positive/emergentFlow.gradle
```diff
@@ -0,0 +1,68 @@
...
+
+// Flow for testing yaml creation for emergent flows for Flow 2.0
+
+hadoop {
+  buildPath "jobs/emergentFlow"
+  cleanPath false
+
+  generateYamlOutput true
+
+  namespace('otherNamespace') {
```
jamiesjc: What would happen with your change if workflow is contained in the namespace block? What if namespace is included in a hadoopClosure and evalHadoopClosures is called?

hadoop-plugin-test/src/main/gradle/positive/emergentFlow.gradle
```diff
@@ -0,0 +1,68 @@
...
+}
+
+hadoopZip {
+  libPath = "lib"
+  base {
+    from { fileTree("jobs/emergentFlow").exclude("otherNamespace").files } {
+      into "."
+    }
+  }
+  zip("emergentFlowTest") {
```
jamiesjc: How about adding two namespace in the test so that it would better demonstrate your fix?

hadoop-plugin-test/src/main/gradle/positive/emergentFlow.gradle
```diff
@@ -0,0 +1,68 @@
...
+    from { fileTree("jobs/emergentFlow/otherNamespace").files } {
+      into "."
+    }
+  }
+}
+
+startHadoopZips.dependsOn buildAzkabanFlows
```
jamiesjc: We usually have one workflow.gradle and one build.gradle file. In this test, it seems you combine them together in one single gradle file. Do they actually work in the same way?

hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/AzkabanDslYamlCompiler.groovy
```diff
@@ -132,6 +129,13 @@ class AzkabanDslYamlCompiler extends BaseCompiler {
       visitNamespace(namespace);
     }
 
+    // Visit properties only if there are no workspaces in the namespace
```
jamiesjc: typo: no workflows?

hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/AzkabanDslYamlCompiler.groovy
```diff
@@ -97,20 +98,16 @@ class AzkabanDslYamlCompiler extends BaseCompiler {
   }
 
   /**
-   * Create and customize a Yaml object.
-   * DumperOptions.FlowStyle.BLOCK indents the yaml in the expected, most readable way.
+   * Instead of visiting workflows and namespaces in the same way the BaseNamedScopeContainer did.
```
jamiesjc: I don't quite follow the comment here. Can you elaborate?

hadoop-plugin/src/main/groovy/com/linkedin/gradle/zip/HadoopZipExtension.groovy
reallocf: Discussed briefly offline. Job file names in Flow 1.0 only contained "_" after being processed by the DSL, users could not declare a job with name "my_test".

This did remind me in the else if block to also check if the file is a .flow file, which apparently I had forgotten. Adding in the next commit - thanks!

hadoop-plugin-test/src/main/gradle/positive/emergentFlow.gradle
reallocf: From my understanding, creating a hadoopClosure and calling evalHadoopClosures is effectively the same thing as creating a namespace.

If a workflow is defined in a namespace block then it will be included at the root of the zip if selected at zip time, then if there are .tempprops files generated in OTHER namespaces are also included in the zip then everything will be zipped together. However, if a properties file is defined in a namespace with the workflow, it will NOT create a .tempprops file because these properties are assumed to be workflow level properties (and so not shared).

hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/AzkabanDslYamlCompiler.groovy
reallocf: Yes - thank you!

hadoop-plugin-test/src/main/gradle/positive/emergentFlow.gradle
reallocf: So I wasn't actually able to get this zip step to build - I think it's a consequence of gradle sequencing (i.e. it should be in the build.gradle file). I'd rather not add the layer of complication there, so I'll probably just remove this step. That means that this test is primarily used to confirm that the .tempprops file is created as expected, not necessarily to test the zipping step. I'm not sure of a good automated way to do this.

hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/AzkabanDslYamlCompiler.groovy
reallocf: Looks like a typo - thanks!

hadoop-plugin/src/main/groovy/com/linkedin/gradle/zip/HadoopZipExtension.groovy
```diff
@@ -172,6 +173,10 @@ class HadoopZipExtension {
       project.tasks["buildHadoopZips"].dependsOn task;
       dependsOn "startHadoopZips";
 
+      doFirst {
+        mergeTempPropsWithFlowFiles(task, zipName);
```
jamiesjc: Is there any test for merging the tempprops with flow YAML files in the zip task?

PR #222, opened 2018-08-17
Add conditional workflow feature.

jamiesjc:
This is part of the conditional workflow design: https://github.com/azkaban/azkaban/issues/1826

We need to generate the condition section for each job or subflow that has user-defined conditions.
Example: Define the condition for JobB in the workflow block.

workflows.gradle
```
workflow('AZTest') {
  commandJob('JobA') {
    uses 'bash ./sample_script.sh'
  }

  commandJob('JobB') {
    uses 'echo "If param1 in JobA is 1, run JobB."'
    depends 'JobA'
    conditions '${JobA:param1} == 1'
   }
  
  targets 'JobB'
}
```

### DIFF
--- a/hadoop-plugin/src/main/groovy/com/linkedin/gradle/hadoopdsl/Workflow.groovy
+++ b/hadoop-plugin/src/main/groovy/com/linkedin/gradle/hadoopdsl/Workflow.groovy
@@ -85,6 +85,9 @@ class Workflow extends BaseNamedScopeContainer {
   // The SubFlowJob for the workflow if this is a grouping subflow
   SubFlowJob subFlowJob;
 
+  // The condition for a subflow to run
+  String condition;
+
   /**
    * Base constructor for a Workflow.
    *
@@ -362,6 +365,11 @@ class Workflow extends BaseNamedScopeContainer {
     boolean clear = args.containsKey("clear") ? args["clear"] : false;
     List<String> targetNames = (List<String>)args["targetNames"];
     flowDepends(clear, targetNames);
+  }
+
+  @HadoopDslMethod
+  void conditions(String subflowCondition) {
+    condition = subflowCondition;
   }
 
   /**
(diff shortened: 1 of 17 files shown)


hadoop-plugin/src/main/groovy/com/linkedin/gradle/hadoopdsl/Workflow.groovy
```diff
@@ -364,6 +367,11 @@ class Workflow extends BaseNamedScopeContainer {
     flowDepends(clear, targetNames);
   }
 
+  @HadoopDslMethod
+  void conditions(String subflowCondition) {
```
reallocf: Should this be `conditions` or `condition`? If you're only able to specify a single condition then making this singular makes more sense to me

hadoop-plugin/src/main/groovy/com/linkedin/gradle/hadoopdsl/job/Job.groovy
```diff
@@ -254,6 +256,11 @@ class Job {
     depends(clear, targetNames);
   }
 
+  @HadoopDslMethod
+  void conditions(String jobCondition) {
```
reallocf: Same as above

hadoop-plugin/src/main/groovy/com/linkedin/gradle/hadoopdsl/Workflow.groovy
reallocf: Hmm, condition is more frequently used as a noun then the other examples you mentioned, but I see what you mean and it probably _can_ be used as a verb in this way.

I think `conditions` should be fine 👍 thanks!

reallocf:
Talked offline about adding additional conditions. @jamiesjc said she'd do this in the next commit.

Also need to update VERSIONS.md and CONTRIBUTORS.md.

jamiesjc:
Thanks for the review @reallocf 
Todo in the next PR:
1. Add WARN checker for workflow root level condition.
2. Add ERROR checker for invalid conditions.

PR #229, opened 2018-09-11
Fix variable substitution issue for dependency list.

jamiesjc:
Similar to the issue in #228 , it will cause an incompatible issue when defining `depends` or `targets` as a list which contains `${}` when generating YAML files.
For example, in workflow.gradle:
```
def jobName = "job1"
def targetList = ["${jobName}"]
hadoop {
  workflow('wordCountFlow') {
    hadoopShellJob("${jobName}") {
      uses "bash ./test.sh"
    }
    targets targetList
  }
}
```
This would throw below error:
```
- name: wordCountFlow
  type: noop
  dependsOn:
  - !!org.codehaus.groovy.runtime.GStringImpl
    metaClass: &id001 !!groovy.lang.MetaClassImpl {}
```
We should also explicitly do the `toString() `conversion for the dependency list.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -156,3 +156,4 @@ The following were contributed by Jamie Sun. Thanks, Jamie!
 * `Add conditional workflow feature`
 * `Fix base properties scope for subflows when generating YAML files`
 * `Fix variable substitution issue when generating YAML files`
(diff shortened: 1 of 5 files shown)


jamiesjc:
I've released the MP locally and verified the fix with the user.

PR #238, opened 2019-01-16
Stop outputting max wait mins when it's defined in a flow trigger with no dependencies

burgerkingeater:
When a flow trigger contains no dependencies, max wait mins won't be required, and the resulting flow yaml file should not contain max wait mins when it's not defined on DSL side.


oliverhu:
You also need to bump version in order to pick this up.

PR #246, opened 2019-03-18
Add xmx information to console error message

edwinalu:
With Gradle 5, the default xmx is very low. Add a message with instructions to set explicitly for newer versions of Gradle, if there is an exception when getting the console.

### DIFF
--- a/hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/AzkabanHelper.groovy
+++ b/hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/AzkabanHelper.groovy
@@ -64,8 +64,11 @@ class AzkabanHelper {
                      |At LinkedIn, you must pass --no-daemon and explicitly set the JAVA_HOME environment variable to
                      |the Java version specified in your product-spec. GRADLE_OPTS environment variable should also be set to
                      |explicitly specify Xmx for Gradle 5 and newer versions, which set a very low default memory threshold for the launcher
-                     |VM; for example: "export GRADLE_OPTS="-Xmx1024m -XX:MaxMetaspaceSize=256m". Refer to the 'Password Masking' section of
-                     |http://go/HadoopPlugin for more info.""".stripMargin();
+                     |VM; for example: "export GRADLE_OPTS="-Xmx1024m -XX:MaxMetaspaceSize=256m".
+                     |Another option for setting Xmx is to set org.gradle.jvmargs in
+                     |gradle.properties. Refer to the 'Password Masking' section of
+                     |https://github.com/linkedin/linkedin-gradle-plugin-for-apache-hadoop/wiki/Azkaban-Features
+                     |for more info.""".stripMargin();
 
   /**
    * Helper method to make a login request to Azkaban.
--- a/hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/AzkabanHelper.groovy
+++ b/hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/AzkabanHelper.groovy
@@ -58,6 +58,15 @@ class AzkabanHelper {
 
   private final static Logger logger = Logging.getLogger(AzkabanHelper);
   static String passwordString = "Password";
+  final static String CONSOLE_EXCEPTION_MESSAGE = """Cannot access the system console. Be sure to 
+pass --no-daemon in your command. For more info,
+                     |refer to https://github.com/linkedin/linkedin-gradle-plugin-for-apache-hadoop/wiki/Azkaban-Features#password-masking.
+                     |
+                     |At LinkedIn, you must pass --no-daemon and explicitly set the JAVA_HOME environment variable to
+                     |the Java version specified in your product-spec. GRADLE_OPTS environment variable should also be set to
+                     |explicitly specify Xmx for Gradle 5 and newer versions, which set a very low default memory threshold for the launcher
+                     |VM; for example: "export GRADLE_OPTS="-Xmx1024m -XX:MaxMetaspaceSize=256m". Refer to the 'Password Masking' section of
+                     |http://go/HadoopPlugin for more info.""".stripMargin();
 
   /**
    * Helper method to make a login request to Azkaban.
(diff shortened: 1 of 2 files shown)


hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/AzkabanHelper.groovy
```diff
@@ -58,6 +58,14 @@ class AzkabanHelper {
 
   private final static Logger logger = Logging.getLogger(AzkabanHelper);
   static String passwordString = "Password";
+  final static String CONSOLE_EXCEPTION_MESSAGE = """Cannot access the system console. Be sure to pass --no-daemon in your command. For more info,
+                     |refer to https://github.com/linkedin/linkedin-gradle-plugin-for-apache-hadoop/wiki/Azkaban-Features#password-masking.
+                     |
+                     |At LinkedIn, you must pass --no-daemon and explicitly set the JAVA_HOME environment variable to
```
li-ygerchikov: I know this isn't new, but any chance we could rephrase this in non-LinkedIn-specific terms? Go-links mean nothing to folks outside LinkedIn.

hadoop-plugin/src/test/groovy/com/linkedin/gradle/azkaban/AzkabanHelperTest.groovy
```diff
@@ -16,10 +16,16 @@
...
-import org.json.JSONObject;
-import org.junit.Test;
+import org.json.JSONObject
+import org.junit.Rule;
+import org.junit.Test
+import org.junit.rules.ExpectedException
+import org.junit.runner.RunWith;
+
 import static org.junit.Assert.assertEquals;
+//import org.powermock.modules.junit4.PowerMockRunner;
```
li-ygerchikov: If these aren't needed, could be get rid of them? Is there a reason to keep commented-out code around?

edwinalu: Removed.

hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/AzkabanHelper.groovy
edwinalu: We could remove the reference completely, but it has more information. Adding the information itself including code snippets may be too much. The repo is already under linkedin.

hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/AzkabanHelper.groovy
```diff
@@ -58,6 +58,14 @@ class AzkabanHelper {
...
+                     |the Java version specified in your product-spec. GRADLE_OPTS environment variable should also be set to
+                     |explicitly specify Xmx for Gradle 5 and newer versions, which set a very low default memory threshold for the launcher
+                     |VM; for example: "export GRADLE_OPTS="-Xmx1024m -XX:MaxMetaspaceSize=256m". Refer to the 'Password Masking' section of
```
erwa: I think we should encourage setting `org.gradle.jvmargs` in `gradle.properties` instead as that seems more self-contained than setting an environment variable every time.

hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/AzkabanHelper.groovy
```diff
@@ -58,6 +58,14 @@ class AzkabanHelper {
...
+                     |http://go/HadoopPlugin for more info.""".stripMargin();
```
erwa: I agree with @li-ygerchikov and think we should update the Wiki documentation (https://github.com/linkedin/linkedin-gradle-plugin-for-apache-hadoop/wiki/User-Guide) and point there instead.

hadoop-plugin/src/test/groovy/com/linkedin/gradle/azkaban/AzkabanHelperTest.groovy
```diff
@@ -31,4 +34,14 @@ class AzkabanHelperTest {
   public void TestGradleException() {
     AzkabanHelper.fetchSortedFlows(new JSONObject("{\"project\" : \"test-azkaban\",   \"projectId\" : 192,   \"flows\" : [ ] }"));
   }
+
+  @Rule
+  public ExpectedException expectedEx = ExpectedException.none();
+
+  @Test(expected = Exception.class)
```
erwa: Wow, I dislike that you have to create a `@Rule ExpectedException` so you can verify the exception message in JUnit. In TestNG, the `@Test` annotation directly supports an `expectedExceptionsMessageRegExp` parameter.

It's also weird to me that you first initialize `expectedEx` to `none()` and then tweak it in the test itself. But I see this is what the JUnit Javadoc's example does, too: https://junit.org/junit4/javadoc/4.12/org/junit/rules/ExpectedException.html

hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/AzkabanHelper.groovy
edwinalu: I'll mention both options.

hadoop-plugin/src/test/groovy/com/linkedin/gradle/azkaban/AzkabanHelperTest.groovy
edwinalu: It is a bit weird, but the tests are using JUnit right now.

erwa:
Thanks for addressing my comments. Going to merge now.

hadoop-plugin/src/main/groovy/com/linkedin/gradle/azkaban/AzkabanHelper.groovy
DPUkyle: @erwa Setting `org.gradle.jvmargs` in `gradle.properties` is "too late". The env var `GRADLE_OPTS` is required to give the launcher VM enough heap to allow builds to run there. When `org.gradle.jvmargs` is set with Gradle 5, you'll still get a single-use daemon spawned in a separate process and that's what breaks console input.

PR #249, opened 2019-05-30
Add a workaround to handle visiting files multiple times when merging tempprops into .flow files.

ypadron-in:
The [visit](https://docs.gradle.org/current/javadoc/org/gradle/api/file/FileTree.html#visit-org.gradle.api.Action-) function from Gradle in HadoopZipExtension.mergeTempPropsWithFlowFiles() visits files multiple times. This causes the error below when trying to create a merge file it already created on a previous visit to the same file and marked it as read only.

```
Caused by: java.io.FileNotFoundException:/Users/ypadron/azkaban/myProject/aFlow.flow (Permission denied)
        at com.linkedin.gradle.zip.YamlMerge.merge(YamlMerge.groovy:42)
        at com.linkedin.gradle.zip.YamlMerge$merge$0.call(Unknown Source)
        at com.linkedin.gradle.zip.HadoopZipExtension$_mergeTempPropsWithFlowFiles_closure4.doCall(HadoopZipExtension.groovy:247)
        ... 55 more
```

Because the root cause of the issue is in code we don’t control the best option is to implement a workaround to handle the multiple visits.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -163,3 +163,6 @@ The following were contributed by Jamie Sun. Thanks, Jamie!
 The following were contributed by Rakesh Malladi. Thanks, Rakesh!
 * `Add ability for Hadoop DSL to understand the WormholePushJob job type`
 * `Deprecate dataset.group property from the WormholePushJob`
+
+The following were contributed by Yeni Bermudez. Thanks, Yeni!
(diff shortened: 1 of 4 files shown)


jamiesjc:
Thanks for the fix. Please also bump gradle.properties to next version and update VERSION.md and CONTRIBUTORS.md files.

PR #251, opened 2019-06-04
Added support for KabootarJob job type to package trained models and …

saikadam:
…make them available for Pro-ML model deployment pipeline.


hadoop-plugin/src/main/groovy/com/linkedin/gradle/hadoopdsl/BaseNamedScopeContainer.groovy
```diff
@@ -25,7 +25,8 @@ import com.linkedin.gradle.hadoopdsl.job.HdfsWaitJob;
 import com.linkedin.gradle.hadoopdsl.job.HiveJob;
 import com.linkedin.gradle.hadoopdsl.job.JavaJob;
 import com.linkedin.gradle.hadoopdsl.job.JavaProcessJob;
-import com.linkedin.gradle.hadoopdsl.job.Job;
+import com.linkedin.gradle.hadoopdsl.job.Job
```
li-ygerchikov: Why remove semicolon?

saikadam: I think, some setting in IDE caused it. Thanks for catching. I added it back.

PR #255, opened 2019-09-09
Refactor AzkabanHelper to avoid System#console calls

DPUkyle:
The `azkabanUpload` task relies on System#console calls to receive credentials. This requires user/environment special behavior to disable the Gradle daemon via the GRADLE_OPTS env var, `--no-daemon` CLI or other means.

This PR replaces that behavior with a better option, `project.ant.input`. This leverages Gradle's bundled ant instance to collect input, and should allow the Gradle daemon to be used for building azkaban zips.

...unfortunately I have no idea how to test this :) I'll get in touch with the owners via other means.


DPUkyle:
This change was untested (and has no test coverage) and in retrospect does not work.

Note that there is no better solution currently; see gradle/gradle#2842

PR #262, opened 2020-01-23
buildMetadata.json: Add the project directory to improve static analysis

SurenNihalani:
Why: helps with LinkedIn's internal migration

What changed: added a projectDirectory which helps us identify the subproject in a MP

Tests performed: modified unit tests


SurenNihalani:
open to other suggestions on what to name this config variable

li-ygerchikov (commented):
Two concerns: (1) project directory may still not tell us much about the MP, and (2) it smells a bit like spying on our users.

SurenNihalani:
1. BuildMetadata.json stores the git remote url for the repo and that’s good enough to get us to a mp name
2. We don’t get to spy on the artifacts so it’s not too strong spying I hope. ideally I’d want Azkaban projects to be defined using infra as code principles (like terraform resources). That’s not the current state so I am settling for static links to projects

li-gsalia (approved):

PR #265, opened 2020-02-27
ScmPlugin: Try to generate buildMetaData.json even when scmPlugin is disabled

SurenNihalani:
People seem to disable scm plugin to improve their zip size/performance.
Changed the flag to two flags to allow metadata.json to be created by
default but for projects where existing flags is set to false, the zip will not be created

### DIFF
--- a/hadoop-plugin/src/main/groovy/com/linkedin/gradle/scm/ScmPlugin.groovy
+++ b/hadoop-plugin/src/main/groovy/com/linkedin/gradle/scm/ScmPlugin.groovy
@@ -27,24 +27,17 @@ import org.gradle.api.tasks.bundling.Zip
 import java.nio.file.Path
 import java.nio.file.Paths;
 
+
 /**
  * ScmPlugin implements features that generate source control management (scm) metadata, in
- * particular for Git and Subversion.
- */
+ * particular for Git and Subversion.*/
 class ScmPlugin implements Plugin<Project> {
-  /**
-   * Applies the ScmPlugin.
-   *
-   * @param project The Gradle project
-   */
-  @Override
-  void apply(Project project) {
-    // Enable users to skip the plugin
-    if (project.hasProperty("disableScmPlugin")) {
-      println("ScmPlugin disabled");
-      return;
-    }
 
+  void applyMetadataTasks(Project project) {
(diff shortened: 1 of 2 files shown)


hadoop-plugin/src/main/groovy/com/linkedin/gradle/scm/ScmPlugin.groovy
```diff
@@ -27,24 +27,17 @@ import org.gradle.api.tasks.bundling.Zip
...
-   */
-  @Override
-  void apply(Project project) {
-    // Enable users to skip the plugin
-    if (project.hasProperty("disableScmPlugin")) {
-      println("ScmPlugin disabled");
-      return;
-    }
 
+  void apply_metadata_tasks(Project project) {
```
ypadron-in: Please use camel case since the rest of the file uses that.

hadoop-plugin/src/main/groovy/com/linkedin/gradle/scm/ScmPlugin.groovy
```diff
@@ -83,7 +76,20 @@ class ScmPlugin implements Plugin<Project> {
...
+  }
+  /**
+   * Applies the ScmPlugin.
+   *
+   * @param project The Gradle project
+   */
+  @Override
+  void apply(Project project) {
+    apply_metadata_tasks(project)
+    // Enable users to skip the plugin to avoid
```
ypadron-in: Can you finish this comment?

hadoop-plugin/src/main/groovy/com/linkedin/gradle/scm/ScmPlugin.groovy
```diff
@@ -83,7 +76,16 @@ class ScmPlugin implements Plugin<Project> {
...
       }
     }
+  }
 
+  @Override
+  void apply(Project project) {
+    applyMetadataTasks(project)
+    // Allows users to skip the plugin if they don't want the sources zip (performance improvement)
+    if (project.hasProperty("disableScmPlugin")) {
+      println("ScmPlugin disabled");
```
ypadron-in: Disabling the plugin before meant no sources zip and no buildMetadata.json file, now will mean no sources zip only.
Wondering if we need to be more specific in the message of this _println_.

